PR #1, opened 2022-10-17
Add OVNDBCluster and OVNNorthd

karelyatin:


api/v1alpha1/ovndbcluster_types.go
```diff
@@ -0,0 +1,164 @@
...
+// OVNDBClusterList contains a list of OVNDBCluster
+type OVNDBClusterList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           []OVNDBCluster `json:"items"`
+}
+
+func init() {
+	SchemeBuilder.Register(&OVNDBCluster{}, &OVNDBClusterList{})
+}
```
mdbooth, line 157: Move this into the parent package, i.e. `api` rather than `api/v1alpha1`.

api/v1alpha1/ovndbcluster_types.go
```diff
@@ -0,0 +1,164 @@
...
+			appsv1.Resource("OVNDBCluster"),
+			fmt.Sprintf("No OVNDBCluster object found in namespace %s", namespace),
+		)
+	}
+	DBEndpointsMap := make(map[string]string)
+	for _, ovndb := range ovnDBList.Items {
+		DBEndpointsMap[ovndb.Spec.DBType] = ovndb.Status.DBAddress
+	}
+	return DBEndpointsMap, nil
+}
```
mdbooth, line 73: Don't include code in the API: move this into some other package.

The way go imports work a user can import just a single package and pull in only the dependencies from that package. Importing the API from another project is a very common pattern. By including code in the api package we dramatically increase the required dependencies of the api package, increasing the probability of diamond dependency issues when importing it from another project.

api/v1alpha1/ovndbcluster_types.go
```diff
@@ -0,0 +1,164 @@
...
+func init() {
+	SchemeBuilder.Register(&OVNDBCluster{}, &OVNDBClusterList{})
+}
+
+// IsReady - returns true if service is ready to server requests
+func (instance OVNDBCluster) IsReady() bool {
+	// Ready when:
+	// there is at least a single pod to serve the OVN DBCluster
+	return instance.Status.ReadyCount >= 1
+}
```
mdbooth, line 164: Ditto.

controllers/ovndbcluster_controller.go
```diff
@@ -0,0 +1,389 @@
...
+func (r *OVNDBClusterReconciler) reconcileNormal(ctx context.Context, instance *ovnv1.OVNDBCluster, helper *helper.Helper) (ctrl.Result, error) {
+	r.Log.Info("Reconciling Service")
+
+	// If the service object doesn't have our finalizer, add it.
+	controllerutil.AddFinalizer(instance, helper.GetFinalizer())
+	// Register the finalizer immediately to avoid orphaning resources on delete
+	//if err := patchHelper.Patch(ctx, openStackCluster); err != nil {
+	if err := r.Update(ctx, instance); err != nil {
+		return ctrl.Result{}, err
+	}
```
mdbooth, line 206: For safety you should return from the reconciler after doing an update. If you don't you can get read-after-write consistency errors in the reconciler (ask me how I know 😬).

The reason is that the reconciler is called with an object from a shared informer. This has a watch and a cache. As soon as you call Update() here the watch is going to update the shared informer, and this reconciler is queued to be called again *with the state of the object at this point*. i.e. Without any of the things we're adding below. We then make all the changes below and when the calling function exits we call our defer func, which updates the object again. At this point there is a race. The shared informer already has a previous reconcile queued for us. If we get called again before kube-apiserver has fired the watch from the second update and the informer has updated based on it, we will be called with old data. i.e. We write new data and then immediately get called again with a version of the object which doesn't contain the changes we just wrote. Worse, this is extremely likely to happen in practise.

Given that we're going to be called again anyway once we call update we might as well exit after a single update call and wait. It's much less likely to result in hair loss.

mdbooth (commented):
Sorry, just a fly-by review with a few generic things. I don't have the context for a full review. Best of luck!

controllers/ovnnorthd_controller.go
```diff
@@ -0,0 +1,371 @@
...
+func (r *OVNNorthdReconciler) reconcileNormal(ctx context.Context, instance *ovnv1.OVNNorthd, helper *helper.Helper) (ctrl.Result, error) {
+	r.Log.Info("Reconciling Service")
+
+	// If the service object doesn't have our finalizer, add it.
+	controllerutil.AddFinalizer(instance, helper.GetFinalizer())
+	// Register the finalizer immediately to avoid orphaning resources on delete
+	//if err := patchHelper.Patch(ctx, openStackCluster); err != nil {
+	if err := r.Update(ctx, instance); err != nil {
+		return ctrl.Result{}, err
+	}
```
abays, line 205: We should address this here and elsewhere where the problematic pattern that Matt mentioned is repeated

controllers/reconciler_common.go
```diff
@@ -31,16 +31,15 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
+// ReconcilerCommon -
 type ReconcilerCommon interface {
 	GetClient() client.Client
 	GetLogger() logr.Logger
 }
```
abays, line 38: Perhaps the content of this `controllers/reconciler_common.go` file is now outdated by equivalent newer options in `lib-common`?

pkg/ovnnorthd/deployment.go
```diff
@@ -0,0 +1,125 @@
...
+	// ServiceCommand -
+	ServiceCommand = "/usr/local/bin/kolla_set_configs && /usr/local/bin/kolla_start"
+)
+
+// Deployment func
+func Deployment(
+	instance *ovnv1.OVNNorthd,
+	configHash string,
+	labels map[string]string,
+) *appsv1.Deployment {
```
abays, line 36: Can we have this use the `lib-common` Deployment wrapper to maintain consistency with other operators?

controllers/ovndbcluster_controller.go
```diff
@@ -0,0 +1,389 @@
...
+func (r *OVNDBClusterReconciler) GetScheme() *runtime.Scheme {
+	return r.Scheme
+}
+
+//+kubebuilder:rbac:groups=ovn.openstack.org,resources=ovndbclusters,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=ovn.openstack.org,resources=ovndbclusters/status,verbs=get;update;patch
+//+kubebuilder:rbac:groups=ovn.openstack.org,resources=ovndbclusters/finalizers,verbs=update
+//+kubebuilder:rbac:groups=core,resources=configmaps,verbs=get;list;watch;create;update;patch;delete;
+//+kubebuilder:rbac:groups=apps,resources=statefulsets,verbs=get;list;watch;create;patch;update;delete;
+//+kubebuilder:rbac:groups=core,resources=services,verbs=get;list;watch;create;update;delete;
```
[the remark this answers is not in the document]
karelyatin: Done

controllers/ovndbcluster_controller.go
karelyatin: Thanks Matthew and Andrew, Done

api/v1alpha1/ovndbcluster_types.go
karelyatin: Moved to seperate package

api/v1alpha1/ovndbcluster_types.go
karelyatin: Actually this part is generated by operator-sdk so am hesitant to move it. AFAIU idea is to register all the resources at single place? and all other openstack operators using similarly. Can you share more context/references around it as i think if that's recommended thing then we can change it in all operators together.

api/v1alpha1/ovndbcluster_types.go
karelyatin: Not moved this method as it not pulling any extra dependencies and just relying on instance Status and will always we used with API object.

controllers/reconciler_common.go
karelyatin: Yes likely, this module will be cleaned up later with other resources.

pkg/ovndbcluster/statefulset.go
```diff
@@ -0,0 +1,132 @@
...
+	// ServiceCommand -
+	ServiceCommand = "/usr/local/bin/kolla_set_configs && /usr/local/bin/kolla_start"
+)
+
+// StatefulSet func
+func StatefulSet(
+	instance *ovnv1.OVNDBCluster,
+	configHash string,
+	labels map[string]string,
+) *appsv1.StatefulSet {
```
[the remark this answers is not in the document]
karelyatin: Using NewStatefulset method in controllers which returns StatefulSet wrapper from lib-common

pkg/ovnnorthd/deployment.go
karelyatin: Using NewDeployment method in controllers which returns Deployment wrapper from lib-common

pkg/ovndbcluster/service.go
```diff
@@ -0,0 +1,42 @@
...
+import (
+	labels "github.com/openstack-k8s-operators/lib-common/modules/common/labels"
+	ovnv1 "github.com/openstack-k8s-operators/ovn-operator/api/v1alpha1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+// Service func
+func Service(instance *ovnv1.OVNDBCluster, scheme *runtime.Scheme) *corev1.Service {
```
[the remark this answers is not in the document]
karelyatin: Done, now using NewService method in controllers which returns Service wrapper from lib-common

controllers/ovndbcluster_controller.go
gibizer: @mdbooth you just described something very similar to my last week head scratching (documented here https://github.com/openstack-k8s-operators/nova-operator/issues/110) where I saw stale data being read at the start of Reconcile. If you know about any similar surprise factors regarding operator development please post it publicly (or point me to those posts if it is already written up somewhere).

abays (approved):
Good first steps, so let's move forward

/lgtm

PR #2, opened 2022-11-01
Create api module and move GetDBEndpoints to it

karelyatin:
GetDBEndpoints will also be used by other operators like neutron, octavia etc.


api/v1alpha1/client.go
```diff
@@ -14,14 +14,13 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package client
+package v1alpha1
```
slawqo, line 17: it's not for this change maybe but AFAIK all other operators starts with v1beta1 so maybe You should change it in ovn-operator too

slawqo (approved):

api/v1alpha1/client.go
karelyatin: Thanks slawek, if that's the case can do that in follow up.

abays (approved):
/lgtm

karelyatin:
Thanks Andrew and Slawek, Merging it and will update neutron-operator PR.

PR #3, opened 2022-11-21
Switch api version to v1beta1

karelyatin:
The targetted version of all the OVN CRDs is v1beta1, renaming it now at this early stage would avoid
disruption in other operators in future as these apis will be used by multiple other operators.


abays (approved):
/lgtm

stuggi (approved):
/lgtm

PR #4, opened 2022-11-22
Raise Error when DBAddress is unset

karelyatin:
OVNNorthd relies on OVNDBCluster Status object
for DBAddress and it is set only when OVNDBCluster is ready. Raise an Error if DBAddress is unset as
without it OVNNorthd or other consumers of OVNDBs
will not work.


abays (approved):
/lgtm

slawqo (approved):

PR #5, opened 2022-11-29
Add VolumeClaimTemplates to ovn db statefulsets

karelyatin:
In order to attach persistent volumes to each pod
of the ovn dbs statefulset add VolumeClaimTemplates. Based on StorageClass and StorageRequest input to the OVNDBCluster CR pvc's will be created and bound to each replica.


pkg/ovndbcluster/statefulset.go
```diff
@@ -113,6 +114,32 @@ func StatefulSet(
...
 	}
+	// https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#persistentvolumeclaim-retention
+	statefulset.Spec.PersistentVolumeClaimRetentionPolicy = &appsv1.StatefulSetPersistentVolumeClaimRetentionPolicy{
+		WhenDeleted: appsv1.DeletePersistentVolumeClaimRetentionPolicyType,
+		WhenScaled:  appsv1.RetainPersistentVolumeClaimRetentionPolicyType,
+	}
+	statefulset.Spec.VolumeClaimTemplates = []corev1.PersistentVolumeClaim{
+		corev1.PersistentVolumeClaim{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "etc-ovn",
```
abays, line 125: Perhaps we should prepend the `instance.Name` to the PVC `Name`?

pkg/ovndbcluster/volumes.go
```diff
@@ -58,6 +58,11 @@ func GetDBClusterVolumeMounts() []corev1.VolumeMount {
 			MountPath: "/var/lib/config-data",
 			ReadOnly:  false,
 		},
+		{
+			Name:      "etc-ovn",
+			MountPath: "/etc/ovn",
+			ReadOnly:  false,
+		},
```
abays, line 65: Does this also require a volume definition or is it able to somehow find the PVC created in the `VolumeClaimTemplate`?

karelyatin: Yes volume definition is not required, just this is enough.

pkg/ovndbcluster/statefulset.go
karelyatin: Yes right should be good to add, will check and update.

karelyatin: Done

abays (approved):
/lgtm

slawqo (approved):

PR #8, opened 2022-12-02
Make sure PVCs get's deleted with OVNDBCluster

karelyatin:
Set Owner reference in PVCs so those get's deleted along with OVNDBCluster CRs.

With Kubernetes v1.23 [alpha] this shouldn't be needed explicitly if PersistentVolumeClaimRetentionPolicy is set to Deleted.

Signed-off-by: Yatin Karel <ykarel@redhat.com>


abays (approved):
/lgtm

PR #9, opened 2022-12-06
Use stable/zed containers as default

karelyatin:
Also set DBType default to "NB" as using
nb container image as default for OVNDBCluster CRD.


abays (approved):
/lgtm

slawqo (approved):

PR #10, opened 2022-12-13
Logging: Use ISO8601 timestamps

karelyatin:
The default value for the zap logger is using epoch timestamps, but that's not user friendly, so we are changing it to using the ISO8601 time format instead.

The zap package from uber is already a dependency from the zap module of the controller-runtime package we already use, so we are not adding any new dependency.

Signed-off-by: Yatin Karel <ykarel@redhat.com>


abays (approved):
/lgtm

PR #11, opened 2022-12-16
Expose OVN North and SouthBound parameters in the API

slawqo:
This patch exposes 4 parameters:

--db-nb-probe-interval-to-active
--db-nb-election-timer
--db-sb-probe-interval-to-active
--db-sb-election-timer

in the API so that cloud operator can configure them. Default values are set to be the same as defaults for the ovn-ctl script (probe-interval-to-active) and as currently used (election-timer).

### DIFF
--- a/api/v1beta1/ovndbcluster_types.go
+++ b/api/v1beta1/ovndbcluster_types.go
@@ -53,6 +53,16 @@ type OVNDBClusterSpec struct {
 	LogLevel string `json:"logLevel"`
 
 	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=10000
+	// OVN Northbound and Southbound RAFT db election timer to use on db creation (in milliseconds)
+	ElectionTimer int32 `json:"electionTimer"`
+
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=60000
+	// Active probe interval from standby to active ovsdb-server remote
+	ProbeIntervalToActive int32 `json:"probeIntervalToActive"`
+
+	// +kubebuilder:validation:Optional
 	// Resources - Compute Resources required by this service (Limits/Requests).
 	// https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/
 	Resources corev1.ResourceRequirements `json:"resources,omitempty"`
--- a/config/crd/bases/ovn.openstack.org_ovndbclusters.yaml
+++ b/config/crd/bases/ovn.openstack.org_ovndbclusters.yaml
@@ -52,6 +52,12 @@ spec:
                 description: DBType - NB or SB
                 pattern: ^NB|SB$
                 type: string
+              electionTimer:
+                default: 10000
+                description: OVN Northbound and Southbound RAFT db election timer
+                  to use on db creation (in milliseconds)
+                format: int32
+                type: integer
               logLevel:
                 default: info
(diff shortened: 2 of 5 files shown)


controllers/ovndbcluster_controller.go
```diff
@@ -387,6 +387,10 @@ func (r *OVNDBClusterReconciler) generateServiceConfigMaps(
 	templateParameters := make(map[string]interface{})
 
 	templateParameters["OVN_LOG_LEVEL"] = instance.Spec.LogLevel
+	templateParameters["OVN_ELECTION_TIMER"] = instance.Spec.ElectionTimer
+	templateParameters["OVN_PROBE_INTERVAL_TO_ACTIVE"] = instance.Spec.ProbeIntervalToActive
+	templateParameters["OVN_ELECTION_TIMER"] = instance.Spec.ElectionTimer
+	templateParameters["OVN_PROBE_INTERVAL_TO_ACTIVE"] = instance.Spec.ProbeIntervalToActive
```
abays, line 393: Is there meant to be some sort of differentiation here?  The keys are duplicated, i.e. `OVN_ELECTION_TIMER` and `OVN_PROBE_INTERVAL_TO_ACTIVE` are used twice.  Should they be given some unique prefix or suffix, like `NB` or `SB` (and then adjust the `ovn-dbcluster-<type>.json` template files to use those unique keys)?

karelyatin: seems just a leftover with the last update, so need cleanup.

config/crd/bases/ovn.openstack.org_ovndbclusters.yaml
```diff
@@ -52,6 +52,12 @@ spec:
                 description: DBType - NB or SB
                 pattern: ^NB|SB$
                 type: string
+              electionTimer:
+                default: 60000
```
karelyatin, line 56: this looks much, isn't 10sec good enough?

controllers/ovndbcluster_controller.go
slawqo: yeah, leftover. Removed :)

config/crd/bases/ovn.openstack.org_ovndbclusters.yaml
slawqo: Yeah, another bad copy-paste :)

api/v1beta1/ovndbcluster_types.go
```diff
@@ -52,6 +52,16 @@ type OVNDBClusterSpec struct {
...
 
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=10000
+	// OVN Northbound and Southbound RAFT db election timer to use on db creation (in milliseconds)
+	ElectionTimer int32 `json:"electionTimer"`
+
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=60000
+	// Active probe interval from standby to active ovsdb-server remote
+	ProbeIntervalToActive int32 `json:"probeInterval"`
```
abays, line 63: I think generally you want to have the JSON key match the name of the field.  i.e. so `json:"probeIntervalToActive"`

slawqo: Done

abays (approved):
/lgtm

PR #15, opened 2023-01-18
Allow attaching service pods to additional network

stuggi:
For network isolation outgoing traffic, service pods need to be attached to the network used by the edpm nodes. To achive this, the `k8s.v1.cni.cncf.io/networks` annotation needs to be set on the pods. This adds a networkAttachment parameter to the CRD where it can be specified to which additional network the service pods should be attached to.

```
spec:
...
      networkAttachment: internalapi
```

Depends-On: openstack-k8s-operators/lib-common#133


stuggi:
for testing quay.io/mschuppe/openstack-operator-index:2cbfda9fc2d54837ef9e4e361402096736b38669 could be used which has references to the required ovs/ovn and also edpm operators.

api/bases/ovn.openstack.org_ovndbclusters.yaml
```diff
@@ -62,6 +66,11 @@ spec:
                 default: info
                 description: LogLevel - Set log level info, dbg, emer etc
                 type: string
+              networkAttachment:
```
stuggi, line 69: would there be a uses case to attach the pods to multiple networks and have another param to specify the one which references the network from which the interface IP should be used for the dbAddress?

karelyatin: > dbAddress

@stuggi Basically for dbAddress we are using [service IPs](https://github.com/openstack-k8s-operators/ovn-operator/blob/main/controllers/ovndbcluster_controller.go#L319-L334)(a k8s service for each POD), so we would like these service IPs to be reachable from EDPM nodes.

stuggi: @karelyatin thats not possible for edpm nodes to be reachable. therefore this PR and what I did in here is that when networkAttachment is set, it will us this network-attachment-definition to add an additional interface to the pods and instead of using the service IPs for the dbAddress it will use the IPs from that network [1]. in setup.sh it will now listen on any interfaces of the pod setting `-db-${DB_TYPE}-addr=0.0.0.0`.
Right now I think its ok to have only a single additional network the pods to attach to, which the edpm nodes can connect to, but want to double check with you.

[1] https://github.com/openstack-k8s-operators/ovn-operator/pull/15/files#diff-4808cc0743c2f09f6c0ee9530436b9b3ef3e64f82a837bcc710d2c8434d67cbcR401

stuggi: the only option would be to use a metallb service with a single vip as loadbalancer, like we do with the api services. this is something I want to test for the longer term. this here was the plan for the march demo

booxter: I don't see why a different component would 1) require access to ovsdb-servers for OVN and 2) not be itself exposed to internalApi network.

@slawqo thoughts?

booxter (approved):
Looks straightforward.

api/bases/ovn.openstack.org_ovndbclusters.yaml
karelyatin: Thanks @stuggi for the details and longer term plans, i basically not checked much the changes in the patch when commented originally. The only concern ariases when using pod IPs do those persist over pod restarts/crashes?

slawqo: @booxter I agree, I don't think there is potentially such use case for other services.

stuggi: @karelyatin the changing IPs is a topic we have to address in a follow up when we work on the IPAM as next step. For now the IPs will most likely change when the pod restarts.

PR #17, opened 2023-01-26
Add pre-commit config

rabi:


.vscode/tasks.json
```diff
@@ -1,6 +1,4 @@
 {
-    // See https://go.microsoft.com/fwlink/?LinkId=733558
-    // for the documentation about the tasks.json format
```
gibizer, line 3: Hm, this is a bit aggressive from the check-json tool even though officially json does not support comments. Btw I'm not really sure the repo needs a .vscode/tasks.json file at all

Makefile
```diff
@@ -270,3 +270,14 @@ golangci: get-ci-tools
...
+
+.PHONY: gowork
+gowork: ## Generate go.work file to support our multi module repository
+	test -f go.work || go work init
+	go work use .
+	go work use ./api
+
+.PHONY: operator-lint
+operator-lint: gowork ## Runs operator-lint
+	GOBIN=$(LOCALBIN) go install github.com/gibizer/operator-lint@latest
```
gibizer, line 282: probably better to pin this to a specific version to avoid CI breakage if a new version is tagged from this that introduce new checks that might fail for the current repo. If latest works then an equivalent would be the last tagged one: `github.com/gibizer/operator-lint@v0.2.1`

pkg/ovndbcluster/volumes.go
```diff
@@ -4,7 +4,8 @@ import corev1 "k8s.io/api/core/v1"
 
 // GetDBClusterVolumes -
 // TODO: merge to GetVolumes when other controllers also switched to current config
-//       mechanism.
+//
+//	mechanism.
```
gibizer, line 8: the extra empty line probably unnecessary if the empty spaces are removed before 'mechanism.'

pkg/ovnnorthd/volumes.go
```diff
@@ -4,7 +4,8 @@ import corev1 "k8s.io/api/core/v1"
 
 // GetNorthdVolumes -
 // TODO: merge to GetVolumes when other controllers also switched to current config
-//       mechanism.
+//
+//	mechanism.
```
gibizer, line 8: ditto

gibizer (changes requested):
I have couple of suggestions inline

.vscode/tasks.json
rabi: Yeah, comments are not allowed like this in json. I would probably leave removal this file for another PR as I don't know the context for it.

Makefile
rabi: Makes sense. I can change this but we do we use latest elsewhere I guess.

pkg/ovndbcluster/volumes.go
rabi: These changes were by the pre-commit hook itself. Changed.

Makefile
gibizer: Yeah when I first introduced operator-lint I used @latest. Then I realized that it means whatever new check is introduced in operator-lint can break our CI until the code is adjusted to the new check. So I tried to go through the repos and pin the version but probably missed some.

pkg/ovndbcluster/volumes.go
gibizer: Yeah, it probably misunderstood the extra spaces before the first word and moved it to a separate paragraph. Thanks for fixing.

gibizer (approved):
look soog

.vscode/tasks.json
amolkahat: May be we can remove this dir and add it to .gitignore.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,53 @@
...
+    - id: destroyed-symlinks
+    - id: check-yaml
+      args: [-m]
+    - id: check-json
+    - id: detect-private-key
+    - id: end-of-file-fixer
+      exclude: ^vendor
+    - id: no-commit-to-branch
+    - id: trailing-whitespace
+      exclude: ^vendor
```
amolkahat, line 53: Adding bashate will be a good choice.

.vscode/tasks.json
rabi: We discussed that earlier. That can be a separate PR, feel free to propose that.

.pre-commit-config.yaml
karelyatin: +1 but that can be a follow up too

karelyatin (approved):

PR #22, opened 2023-02-10
Return error in ConfigMap creation

kajinamit:
The current logic ignores the error but it should be returned so that the error can be detected in the reconciler.

### DIFF
--- a/controllers/ovndbcluster_controller.go
+++ b/controllers/ovndbcluster_controller.go
@@ -467,12 +467,7 @@ func (r *OVNDBClusterReconciler) generateServiceConfigMaps(
 			ConfigOptions: templateParameters,
 		},
 	}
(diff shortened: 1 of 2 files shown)


controllers/ovndbcluster_controller.go
```diff
@@ -467,12 +467,7 @@ func (r *OVNDBClusterReconciler) generateServiceConfigMaps(
 			ConfigOptions: templateParameters,
 		},
 	}
-	err := configmap.EnsureConfigMaps(ctx, h, instance, cms, envVars)
-	if err != nil {
-		return nil
-	}
-
-	return nil
+	return configmap.EnsureConfigMaps(ctx, h, instance, cms, envVars)
```
slawqo, line 470: no it matches with what it should returns :) Thx.

slawqo (approved):

PR #29, opened 2023-03-02
skip GetNADWithName() if no NetworkAttachment is specified

stuggi:


controllers/ovndbcluster_controller.go
```diff
@@ -228,24 +228,26 @@ func (r *OVNDBClusterReconciler) reconcileNormal(ctx context.Context, instance *
...
-		instance.Status.Conditions.Set(condition.FalseCondition(
-			condition.NetworkAttachmentsReadyCondition,
-			condition.ErrorReason,
-			condition.SeverityWarning,
-			condition.NetworkAttachmentsReadyErrorMessage,
-			err.Error()))
-		return ctrl.Result{}, err
 	}
 
 	serviceAnnotations, err := nad.CreateNetworksAnnotation(instance.Namespace, []string{instance.Spec.NetworkAttachment})
```
slawqo, line 253: shouldn't this also be optional? With Your patch I still have errors like:

```
Generated from multus
cannot find a network-attachment-definition () in namespace (openstack): resource name may not be empty
```

In the pod's events list.

stuggi: you are right, missed that. will push an update

karelyatin (approved):
/lgtm

PR #30, opened 2023-03-13
Fix pre-commit findings

gibizer:


slawqo:
I think that this is already fixed by https://github.com/openstack-k8s-operators/ovn-operator/pull/28 so we should be good to close this PR

gibizer:
@slawqo  thanks for noticing this

PR #32, opened 2023-03-22
Expose both Pod's and NAD's IPs as DBAddresses

slawqo:
Both IP addresses may be needed as ovn-controller running inside POD (OVS) on control plane should use POD's network to connect to the OVN DBs and EDPM nodes should use internalapi network which is plugged as NAD to the OVN DB pods.


karelyatin (approved):
/lgtm

PR #37, opened 2023-04-06
Webhook for default images

abays:
Default container images are now handled via env vars and internal webhooks. This defaulting logic will in turn later be integrated with OpenStack Operator via webhooks there that use the new defaulting mechanism (while at the same time disabling the OVN service operator internal webhooks in that context).

Note that at this time, we are only covering `OVNDBCluster` and `OVNNorthd` CRDs with these webhooks, because these are the only CRDs from the OVN group that are deployments are currently using.  If other CRDs from this operator are put into service, we can follow-up by adding defaulting webhooks for those as well.


karelyatin (approved):
/lgtm

PR #39, opened 2023-04-17
Drop Legacy CRDs/Controllers

karelyatin:
These exists from cnosp time and are not
maintained, so let's drop these.


slawqo:
Thanks for this cleaning :)

slawqo (approved):

PR #41, opened 2023-04-19
Add Functional Tests

karelyatin:


tests/functional/ovndbcluster_controller_test.go
```diff
@@ -0,0 +1,277 @@
...
+var _ = Describe("OVNDBCluster controller", func() {
+
+	var namespace string
+	var OVNDBClusterName types.NamespacedName
+
+	BeforeEach(func() {
+		// NOTE(gibi): We need to create a unique namespace for each test run
+		// as namespaces cannot be deleted in a locally running envtest. See
+		// https://book.kubebuilder.io/reference/envtest.html#namespace-usage-limitation
+		namespace = uuid.New().String()
```
mrKisaoLamb, line 43: Namespace can be moved to global BeforeEach function, like it was done in nova-operator: https://github.com/openstack-k8s-operators/nova-operator/blob/master/test/functional/suite_test.go#L199.

tests/functional/ovndbcluster_controller_test.go
```diff
@@ -0,0 +1,277 @@
...
+		th.CreateNamespace(namespace)
+		// We still request the delete of the Namespace to properly cleanup if
+		// we run the test in an existing cluster.
+		DeferCleanup(th.DeleteNamespace, namespace)
+		// lib-common uses OPERATOR_TEMPLATES env var to locate the "templates"
+		// directory of the operator. We need to set them otherwise lib-common
+		// will fail to generate the ConfigMap as it does not find ovn-dbcluster.json
+		err := os.Setenv("OPERATOR_TEMPLATES", "../../templates")
```
mrKisaoLamb, line 51: It would be better to set these values dynamically during the make test, similar to what is done in the nova-operator Makefile here: https://github.com/openstack-k8s-operators/nova-operator/blob/master/Makefile#L139, in order to ensure that the path is correct during runtime (we hit some issues with correctnes of path in nova-operator)

tests/functional/ovndbcluster_controller_test.go
```diff
@@ -0,0 +1,277 @@
...
+
+		It("should have a finalizer", func() {
+			// the reconciler loop adds the finalizer so we have to wait for
+			// it to run
+			Eventually(func() []string {
+				return GetOVNDBCluster(OVNDBClusterName).Finalizers
+			}, timeout, interval).Should(ContainElement("OVNDBCluster"))
+		})
+
+		It("should not create a config-data config map", func() {
```
mrKisaoLamb, line 83: This looks like the script and config-data checks can be combined into one table test, for example, as shown here: https://github.com/openstack-k8s-operators/nova-operator/blob/master/test/functional/nova_reconfiguration_test.go#L324

mrKisaoLamb (commented):
Nice! It's great to see more tests!

mrKisaoLamb:
To me, it looks nice; however, it may require approval from one of the owners

slawqo (approved):

PR #47, opened 2023-04-27
Differentiate NB and SB default images

abays:
The initial defaulting webhooks for OVN did not properly distinguish between container images for NB and SB DB clusters.  This corrects that.


karelyatin (commented):
/lgtm

PR #48, opened 2023-05-04
Adapt to split test helpers in lib-common

gibizer:
The test module in lib-common has been split in
https://github.com/openstack-k8s-operators/lib-common/pull/252 . The new modules/test-operators module the one that has the full set of helpers we need to use in our tests. The previously used modules/test is reduced to only contain non openstack-k8s-operators specific helpers.

This patch simply change the import to use the new module


karelyatin (approved):
/lgtm

PR #62, opened 2023-05-30
GetDBEndpoints: handle case of len(dbClusters) == 1

booxter:
If there's a single dbCluster, immediately return error.


karelyatin (approved):
/lgtm

PR #64, opened 2023-06-01
Fix the defaulting logic for OVNController.spec.external-ids

booxter:
This patch allows to post an empty `spec` section and get reasonable defaults set for external-ids used by ovn-controller.

### DIFF
--- a/api/v1beta1/ovncontroller_types.go
+++ b/api/v1beta1/ovncontroller_types.go
@@ -38,6 +38,7 @@ const (
 // OVNControllerSpec defines the desired state of OVNController
 type OVNControllerSpec struct {
 	// +kubebuilder:validation:Optional
+	// +kubebuilder:default={}
 	ExternalIDS OVSExternalIDs `json:"external-ids"`
 
 	// +kubebuilder:validation:Required
@@ -132,14 +133,18 @@ func (instance OVNController) IsReady() bool {
 
 // OVSExternalIDs is a set of configuration options for OVS external-ids table
 type OVSExternalIDs struct {
-	SystemID  string `json:"system-id"`
-	OvnBridge string `json:"ovn-bridge"`
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default="random"
+	SystemID  string `json:"system-id,omitempty"`
+
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default="br-int"
+	OvnBridge string `json:"ovn-bridge,omitempty"`
 
 	// +kubebuilder:validation:Optional
 	// +kubebuilder:default="geneve"
 	// +kubebuilder:validation:Enum={"geneve","vxlan"}
-	// OvnEncapType - geneve or vxlan
-	OvnEncapType string `json:"ovn-encap-type"`
+	OvnEncapType string `json:"ovn-encap-type,omitempty"`
 
 	// +kubebuilder:validation:Optional
 	// +kubebuilder:default=true
--- a/Makefile
+++ b/Makefile
@@ -314,10 +314,14 @@ gowork: ## Generate go.work file to support our multi module repository
 	go work use .
 	go work use ./api
 
+# Skip C003: "Field 'XXX' has both a 'Optional' kubebuilder marker with a
+# default value and an 'omitempty' tag. Either remove the default value or
+# remove 'omitempty'" since spec.external-ids struct in OVNController CRD
+# relies on both being present to fill in the defaults.
 .PHONY: operator-lint
 operator-lint: gowork ## Runs operator-lint
-	GOBIN=$(LOCALBIN) go install github.com/gibizer/operator-lint@v0.3.0
-	go vet -vettool=$(LOCALBIN)/operator-lint ./... ./api/...
+	GOBIN=$(LOCALBIN) go install github.com/gibizer/operator-lint@v0.4.0
+	go vet -vettool=$(LOCALBIN)/operator-lint -C003.skip ./... ./api/...
 
 # Used for webhook testing
 # Please ensure the ovn-controller-manager deployment and
--- a/api/bases/ovn.openstack.org_ovncontrollers.yaml
+++ b/api/bases/ovn.openstack.org_ovncontrollers.yaml
@@ -53,19 +53,17 @@ spec:
                     default: true
                     type: boolean
                   ovn-bridge:
+                    default: br-int
                     type: string
                   ovn-encap-type:
                     default: geneve
-                    description: OvnEncapType - geneve or vxlan
                     enum:
                     - geneve
                     - vxlan
                     type: string
                   system-id:
+                    default: random
                     type: string
-                required:
-                - ovn-bridge
-                - system-id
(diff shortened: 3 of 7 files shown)


karelyatin (commented):
Thanks Ihar for the patch,  I too noticed it while adding functional tests[1] and didn't find a way apart from duplicating the defaults in ExternalIDS and left it for later.
So while you fix operator-lint issues(use of omitempty with defaults defined) please also clear the external-ids in [1] and can also add the test for default check for these at [2]

[1] https://github.com/openstack-k8s-operators/ovn-operator/blob/main/tests/functional/base_test.go#L275
[2] https://github.com/openstack-k8s-operators/ovn-operator/blob/main/tests/functional/ovncontroller_controller_test.go#L48

api/v1beta1/ovncontroller_types.go
```diff
@@ -119,14 +120,18 @@ func (instance OVNController) IsReady() bool {
...
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default="br-int"
+	OvnBridge string `json:"ovn-bridge,omitempty"`
 
 	// +kubebuilder:validation:Optional
 	// +kubebuilder:default="geneve"
 	// +kubebuilder:validation:Enum={"geneve","vxlan"}
-	// OvnEncapType - geneve or vxlan
-	OvnEncapType string `json:"ovn-encap-type"`
+	OvnEncapType string `json:"ovn-encap-type,omitempty"`
```
gibizer, line 134: I would try to avoid omitempty if possible due to https://github.com/gibizer/operator-lint/tree/main/linters/crd/C003 

Unfortunately I only have a solution that needs some level of repetition of the default value:
https://github.com/gibizer/ovn-operator/commit/0c490b8305095ac7be97507249e2a0751c50ccb2

Interestingly it is enough to repeat one of the field defaults on the struct field level to get the rest automatically and it does not matter which one. (I still hate golang + kubebuilder json serialization logic)

booxter:
@gibizer ok this approach doesn't seem to work: I now have to set `spec: exernal-ids: {}` in my sample files to apply it (which was not needed with my solution). Thoughts?

gibizer:
> @gibizer ok this approach doesn't seem to work: I now have to set `spec: exernal-ids: {}` in my sample files to apply it (which was not needed with my solution). Thoughts?

Interesting. I tested it with a fully empty spec in envtest and the defaulting seemed to work. Let me try again now with your patch (even though I don't see any difference).

gibizer:
> > @gibizer ok this approach doesn't seem to work: I now have to set `spec: exernal-ids: {}` in my sample files to apply it (which was not needed with my solution). Thoughts?
> 
> Interesting. I tested it with a fully empty spec in envtest and the defaulting seemed to work. Let me try again now with your patch (even though I don't see any difference).

Ohh so you took over the envtest test case as well and that still passes. Then I guess you tried in a real crc env. Did you run `make install` before you started this version of the operator in crc?

gibizer:
> > > @gibizer ok this approach doesn't seem to work: I now have to set `spec: exernal-ids: {}` in my sample files to apply it (which was not needed with my solution). Thoughts?
> > 
> > 
> > Interesting. I tested it with a fully empty spec in envtest and the defaulting seemed to work. Let me try again now with your patch (even though I don't see any difference).
> 
> Ohh so you took over the envtest test case as well and that still passes. Then I guess you tried in a real crc env. Did you run `make install` before you started this version of the operator in crc?

Now I tried it in crc. It works for me. I changed the OpenStackControlPlane sample:
```
$ git diff
diff --git a/config/samples/core_v1beta1_openstackcontrolplane_network_isolation.yaml b/config/samples/core_v1beta1_openstackcontrolplane_network_isolation.yaml
index fdaf766..fb93080 100644
--- a/config/samples/core_v1beta1_openstackcontrolplane_network_isolation.yaml
+++ b/config/samples/core_v1beta1_openstackcontrolplane_network_isolation.yaml
@@ -132,10 +132,6 @@ spec:
       ovnNorthd:
         networkAttachment: internalapi
       ovnController:
-        external-ids:
-          system-id: "random"
-          ovn-bridge: "br-int"
-          ovn-encap-type: "geneve"
         networkAttachment: tenant
   placement:
     template:
```

When I deployed that I got 
```
$ oc get OVNController -o yaml
apiVersion: v1
items:
- apiVersion: ovn.openstack.org/v1beta1
  kind: OVNController
  metadata:
    creationTimestamp: "2023-06-07T12:52:40Z"
    finalizers:
    - OVNController
    generation: 1
    name: ovncontroller
    namespace: openstack
    ownerReferences:
    - apiVersion: core.openstack.org/v1beta1
      blockOwnerDeletion: true
      controller: true
      kind: OpenStackControlPlane
      name: openstack-network-isolation
      uid: 3ecb4e8b-3a2a-45d2-96dd-3e0dccedd46c
    resourceVersion: "629263"
    uid: c96aac12-9700-47c0-b5a8-eecd3c78b6f9
  spec:
    external-ids:
      enable-chassis-as-gateway: true
      ovn-bridge: br-int
      ovn-encap-type: geneve
      system-id: random
    networkAttachment: tenant
    ovnContainerImage: quay.io/podified-antelope-centos9/openstack-ovn-controller:current-podified
    ovsContainerImage: quay.io/podified-antelope-centos9/openstack-ovn-base:current-podified
    resources: {}
```

gibizer:
> @gibizer here's the patch to test suite that triggers the kuttl failure: [36e5f79](https://github.com/openstack-k8s-operators/ovn-operator/commit/36e5f797849797bee4e4b465f3c011d6149b225d) With the omitempty back (also pushed in the branch for now), it passes; with the default={...} solution, it doesn't.

Thanks. You are right. If the spec field is also removed then we are back to golang defaults in the nested struct. :/  I gave up. We have to go with the omitempty and we have to be very careful that this means the nested fields cannot be explicitly set to golang default values any more as that value is omitted. If this becomes a problem for a field then we need to switch that field to be a pointer type.

booxter:
@gibizer thanks for confirming! AFAIU we need to adjust the linter first then, right? Is adjusting it to support some "ignore" keyword for a rule the way to go? Anything else?

gibizer:
> @gibizer thanks for confirming! AFAIU we need to adjust the linter first then, right? Is adjusting it to support some "ignore" keyword for a rule the way to go? Anything else?

I've tagged operator-lint v0.4.0 that supports skipping checks. With the following you can skip C003:
```diff
-       GOBIN=$(LOCALBIN) go install github.com/gibizer/operator-lint@v0.3.0
-       go vet -vettool=$(LOCALBIN)/operator-lint ./... ./api/...
+       GOBIN=$(LOCALBIN) go install github.com/gibizer/operator-lint@v0.4.0
+       go vet -vettool=$(LOCALBIN)/operator-lint -C003.skip ./... ./api/...
 ```

booxter:
@karelyatin @slawqo please review, I finally refreshed it with the new linter that allows to skip C003 check. Thanks!

karelyatin (approved):
/lgtm

PR #66, opened 2023-06-07
set enable-chassis-as-gw on controllers by default

luis5tb:
As part of [1] the enable-chassis-as-gw flag has been removed by default from the computes. Now this patch changes the default on the controllers to have it enabled so that the gw ports as hosted on the controllers instead of the computes by default.

[1] https://github.com/openstack-k8s-operators/edpm-ansible/pull/178

### DIFF
--- a/api/v1beta1/ovncontroller_types.go
+++ b/api/v1beta1/ovncontroller_types.go
@@ -128,7 +128,9 @@ type OVSExternalIDs struct {
 	// OvnEncapType - geneve or vxlan
 	OvnEncapType string `json:"ovn-encap-type"`
 
-	EnableChassisAsGateway bool `json:"enable-chassis-as-gateway,omitempty" optional:"true"`
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=true
+	EnableChassisAsGateway bool `json:"enable-chassis-as-gateway"`
 }
 
 // RbacConditionsSet - set the conditions for the rbac object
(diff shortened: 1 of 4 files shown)


karelyatin:
Would also need to validate if this works as expected before we merge it.

luis5tb:
> Would also need to validate if this works as expected before we merge it.

Yep, lets wait until we see the controllers are configure properly to enable N/S traffic, so that configuring the gw ports on the controllers does not break it

booxter (commented):
We should also add a default (true) to CRD definition, no?

luis5tb:
> We should also add a default (true) to CRD definition, no?

Isn't that handled on the init.sh? if no value is passed it will take true (which made me realize that it should be "true" instead of true)

api/v1beta1/ovncontroller_types.go
```diff
@@ -128,7 +128,9 @@ type OVSExternalIDs struct {
 	// OvnEncapType - geneve or vxlan
 	OvnEncapType string `json:"ovn-encap-type"`
 
-	EnableChassisAsGateway bool `json:"enable-chassis-as-gateway,omitempty" optional:"true"`
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=true
+	EnableChassisAsGateway bool `json:"enable-chassis-as-gateway,omitempty"`
```
karelyatin, line 133: current operator-linters don't allow using omitempty with default value assigned, can you drop that?

karelyatin (commented):
/lgtm

karelyatin (approved):

PR #73, opened 2023-06-13
Set ovn-encap-ip correctly with network isolation

karelyatin:
It was fixed in [1] but during [2] it got
reverted.

Since now Job pods are used for the configuration, IP is also used from job pod not from the ovn-controller pod.

Moved setting ip bits into the ovn-controller pod
itself as network is setup during the pod initilization.

[1] https://github.com/openstack-k8s-operators/ovs-operator/pull/75
[2] https://github.com/openstack-k8s-operators/ovs-operator/pull/35


templates/ovncontroller/bin/start-ovsdb-server.sh
```diff
@@ -15,6 +15,8 @@
 # under the License.
 set -ex
 
+source $(dirname $0)/functions
+
```
booxter, line 19: I assume I'm missing something, but how is this functions module used here? I don't see a single thing that refers to any of functions / variables defined there.

karelyatin: yes right, my bad. Will cleanup

booxter (approved):

PR #75, opened 2023-06-15
Fix OVNController ConfigJob waiting

booxter:
The timeout argument is in nanoseconds, so we should multiply it by time.Second to wait 5 seconds instead of 5 nanoseconds.


karelyatin (approved):
/lgtm

Ohkk so it was missed in switch config to
job patch and required since https://github.com/openstack-k8s-operators/lib-common/commit/9327b47a79fd452dc1f2bb012d1b26abb241410b

PR #76, opened 2023-06-15
Don't run northd and ovsdb-server services as root

booxter:
OVNController is a bit more tricky because it uses DaemonSets which apparently don't support PersistentClaims, and HostPaths don't support fsGroups. We'll have to find a way to perhaps switch from DaemonSet first to do the same for OVNController.


config/rbac/role.yaml
```diff
@@ -218,6 +218,14 @@ rules:
   - list
   - update
   - watch
+- apiGroups:
```
karelyatin, line 221: Duplicate of L229

karelyatin (changes requested):
Just one comment inline

config/rbac/role.yaml
booxter: not really? L229+ is anyuid+privileged, still used by OVNController. This file is auto-updated by operator make hooks.

booxter: I updated the title here + clarified in commit messages what's actually happening here + split the single commit into three, separating concerns. Let me know if this helps.

karelyatin: Thanks Ihar, yes right these are generated from annotations defined in *_controllers.go

Still it's doubtful if the definitions from this file is really used since the introduction of Dynamic roles/serviceaccounts, can check with @dprince on this.

karelyatin: > Still it's doubtful if the definitions from this file is really used since the introduction of Dynamic roles/serviceaccounts, can check with @dprince on this.

Ok checked this in a test patch https://github.com/openstack-k8s-operators/ovn-operator/pull/80 and those are needed to create dynamic role/service_accounts. A service account/user can create role only if it had the same or higher permissions https://kubernetes.io/docs/reference/access-authn-authz/rbac/#restrictions-on-role-creation-or-update

PR #78, opened 2023-06-19
[envtest]Common helper for Deployment and StatefulSet

gibizer:
The https://github.com/openstack-k8s-operators/lib-common/pull/277 added the PVC fix to the common helper so that can be used in OVN too.


karelyatin (approved):
/lgtm

PR #80, opened 2023-06-27
[DNM] Check Drop duplicate rules

karelyatin:
These are created dynamically now.


karelyatin:
These Clusterrole rules are required to create dynamic roles https://kubernetes.io/docs/reference/access-authn-authz/rbac/#restrictions-on-role-creation-or-update

PR #84, opened 2023-07-11
Update test helper references after reorg

gibizer:
Depends-On: openstack-k8s-operators/lib-common#301


karelyatin:
sorry @gibizer for duplicating it, can you check https://github.com/openstack-k8s-operators/ovn-operator/pull/83 as that includes changes in suite_test.go? If you prefer to update this then i will abandon mine.

gibizer:
Abandoning in favor of #83

PR #87, opened 2023-07-13
Modify kuttl tests to run in SNO with many nodes

cescgina:
I've thought of two different ways to change the kuttl tests for the
OVNController to work in a SNO cluster with more than one node:
- Don't check the number of pods at all, just check the conditions of the
OVNController CR, and assume that if everything is Ready, it should be good
- Add a script step that queries the number of OVNController pods and compares
it to the number of nodes in the cluster

The two approaches are puhsed as different commits, so we can pick any of the
two, or both and squash the commits.


tests/kuttl/common/assert_sample_deployment.yaml
```diff
@@ -5,14 +5,15 @@
 # - 2 OVNDBCluster CR (ovndbcluster-nb-sample, ovndbcluster-sb-sample)
 # - 1 OVNController CR
 # - Deployment with 1 Pod for OVNNorthd CR
-# - Daemonset with 3 Pods for OVNController CR
+# - Daemonset with for OVNController CR
```
booxter, line 8: with for -> for

tests/kuttl/common/assert_sample_deployment.yaml
```diff
@@ -5,14 +5,15 @@
...
 # - ovsdbserver-nb-0 Pod
 # - ovsdbserver-sb-0 Pod
 # - ovn-controller- Pod
 # - ovsdbserver-nb Service
 # - ovsdbserver-nb-0 Service
 # - ovsdbserver-sb Service
 # - ovsdbserver-sb-0 Service
+# - OVNController Daemonset crated as many pods as nodes
```
booxter, line 16: crated -> created

tests/kuttl/common/assert_sample_deployment.yaml
```diff
@@ -73,8 +74,39 @@ spec:
   ovnContainerImage: quay.io/podified-antelope-centos9/openstack-ovn-controller:current-podified
   ovsContainerImage: quay.io/podified-antelope-centos9/openstack-ovn-base:current-podified
 status:
-  desiredNumberScheduled: 1
-  numberReady: 1
+  conditions:
+  - message: Setup complete
```
booxter, line 78: that's fine though I wonder if we have to check all of the conditions; I think Ready would be enough (I don't know if kuttl will allow to check just the important one though, and there's no significant harm in checking more, so feel free to ignore).

booxter (approved):

tests/kuttl/common/assert_sample_deployment.yaml
cescgina: thanks, fixed

tests/kuttl/common/assert_sample_deployment.yaml
cescgina: thanks, fixed

tests/kuttl/common/assert_sample_deployment.yaml
cescgina: Since this is a list, kuttl needs taht all conditions are there, otherwise it compares the list length and fails. We could however remove fields, like the `message` or `reason` to make it less verbose

karelyatin (commented):
/lgtm

karelyatin (approved):

PR #104, opened 2023-08-01
Create ConfigMap for EDP consumption

booxter:
This is just a draft. Will need a lot of polishing before merging.

### DIFF
--- a/controllers/ovncontroller_controller.go
+++ b/controllers/ovncontroller_controller.go
@@ -25,6 +25,7 @@ import (
 	"github.com/go-logr/logr"
 	netattdefv1 "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1"
 	k8s_errors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/kubernetes"
 	ctrl "sigs.k8s.io/controller-runtime"
@@ -432,9 +433,39 @@ func (r *OVNControllerReconciler) reconcileNormal(ctx context.Context, instance 
 	}
 	// create DaemonSet - end
 
+	sbCluster, err := v1beta1.GetDBClusterByType(ctx, helper, instance.Namespace, map[string]string{}, v1beta1.SBDBType)
+	if err != nil {
+		r.Log.Info("No SB OVNDBCluster defined, deleting external ConfigMap")
+		cleanupConfigMapErr := r.deleteExternalConfigMaps(ctx, helper, instance)
+		if cleanupConfigMapErr != nil {
+			r.Log.Error(cleanupConfigMapErr, "Failed to delete external ConfigMap")
+			return ctrl.Result{}, cleanupConfigMapErr
+		}
+		return ctrl.Result{}, nil
+	}
+
+	_, err = sbCluster.GetExternalEndpoint()
+	if err != nil {
+		r.Log.Info("No external endpoint defined for SB OVNDBCluster, deleting external ConfigMap")
+		cleanupConfigMapErr := r.deleteExternalConfigMaps(ctx, helper, instance)
+		if cleanupConfigMapErr != nil {
+			r.Log.Error(cleanupConfigMapErr, "Failed to delete external ConfigMap")
+			return ctrl.Result{}, cleanupConfigMapErr
+		}
+		return ctrl.Result{}, nil
+	}
+
+	// Create ConfigMap for external dataplane consumption
+	// TODO(ihar) - is there any hashing mechanism for EDP config? do we trigger deploy somehow?
+	err = r.generateExternalConfigMaps(ctx, helper, instance, sbCluster, &configMapVars)
+	if err != nil {
+		r.Log.Error(err, "Failed to generate external ConfigMap")
+		return ctrl.Result{}, err
+	}
+
 	// create OVN Config Job - start
 	if instance.Status.NumberReady == instance.Status.DesiredNumberScheduled {
-		jobsDef, err := ovncontroller.ConfigJob(ctx, helper, r.Client, instance, serviceLabels)
+		jobsDef, err := ovncontroller.ConfigJob(ctx, helper, r.Client, instance, sbCluster, serviceLabels)
 		if err != nil {
 			r.Log.Error(err, "Failed to create OVN controller configuration Job")
 			return ctrl.Result{}, err
@@ -491,7 +522,7 @@ func (r *OVNControllerReconciler) reconcileNormal(ctx context.Context, instance 
 	return ctrl.Result{}, nil
 }
 
-// generateServiceConfigMaps - create create configmaps which hold scripts and service configuration
+// generateServiceConfigMaps - create configmaps which hold scripts and service configuration
 func (r *OVNControllerReconciler) generateServiceConfigMaps(
 	ctx context.Context,
 	h *helper.Helper,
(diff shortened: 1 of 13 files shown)


booxter:
Example `ConfigMap`:

```
[ihrachys@ansible-fx2-6 ovn-operator]$ oc get configmap -o yaml ovn-controller-config
apiVersion: v1
data:
  ovn-encap-type: geneve
  ovn-of-probe-interval: "60000"
  ovn-remote: tcp:10.217.5.248:6642
  ovn-remote-probe-interval: "60000"
kind: ConfigMap
metadata:
  creationTimestamp: "2023-08-01T15:36:41Z"
  labels:
    ovn-controller.openstack.org/name: ovncontroller-sample
    ovn-controller.openstack.org/namespace: openstack-operators
    ovn-controller.openstack.org/uid: ab904250-0a32-4db6-9ce9-7ceea1499d17
  name: ovn-controller-config
  namespace: openstack-operators
  ownerReferences:
  - apiVersion: ovn.openstack.org/v1beta1
    blockOwnerDeletion: true
    controller: true
    kind: OVNController
    name: ovncontroller-sample
    uid: ab904250-0a32-4db6-9ce9-7ceea1499d17
  resourceVersion: "395243"
  uid: 9fa794ac-e6f0-4154-9a49-6fc195d73bf5
```

controllers/ovncontroller_controller.go
```diff
@@ -263,11 +264,45 @@ func (r *OVNControllerReconciler) reconcileNormal(ctx context.Context, instance
...
+	// TODO(ihar) introduce a constructor for internal-SB string
+	externalConfig := make(map[string]string)
+	// TODO(ihar) should we fail on missing key? perhaps the interface of GetDbEndpoints should be more stringent
+	val, ok := dbmap["internal-SB"]
+	if ok {
+		// TODO(ihar) define keys here and elsewhere in a const file
+		externalConfig["ovn-remote"] = val
+	}
+
+	// TODO(ihar) do we need ovn- prefixes for all of the settings?
```
slawqo, line 291: those are just internal keys used in this operator, nothing public, right?
If so, I think we can get rid of the "ovn-" prefix there.

booxter: These are actually - in a way - the public interface for external tools (in case of EDP this is Ansible) that would like to integrate with service operators. From this perspective, I consider them public.

The question about prefix is more about "cleanliness" and avoiding duplication (since the `ConfigMap` name itself contains the `ovn-` prefix). If a tool wants to have `ovn` as part of their path to configuration options, it can always mount the `ConfigMap` to a directory that has `ovn` in its full path.

booxter: While we are thinking in terms of tight coupling between EDPM Ansible and service operators, the fact that this is an interface is of little relevance. What I'm thinking of is - a potential future where EDPM Ansible is no longer used, perhaps because there's a better / alternative way to deploy Compute nodes (as OCP nodes?) In this case, this interface will become shared / reused. This is e.g. why I don't set the keys in the `data` section of the spec with the names used by EDPM Ansible roles (with `edpm_` prefixes etc.) - I don't want this interface to assume a particular way of deploying Compute nodes.

controllers/ovncontroller_controller.go
```diff
@@ -263,11 +264,45 @@ func (r *OVNControllerReconciler) reconcileNormal(ctx context.Context, instance
...
+		// TODO(ihar) is it a valid return while waiting for dbcluster?
+		return ctrl.Result{}, err
+	}
+	// TODO(ihar) check that sbCluster is not nil (or change the interface of the ByType function)
+
+	// TODO(ihar) move the code below to a separate helper function
+	// TODO(ihar) introduce a constructor for internal-SB string
+	externalConfig := make(map[string]string)
+	// TODO(ihar) should we fail on missing key? perhaps the interface of GetDbEndpoints should be more stringent
+	val, ok := dbmap["internal-SB"]
```
karelyatin, line 285: Considering the config map would be used for EDP, we would need to expose dbmap["SB"] instead of dbmap["internal-SB"]

controllers/ovncontroller_controller.go
```diff
@@ -263,11 +264,45 @@ func (r *OVNControllerReconciler) reconcileNormal(ctx context.Context, instance
...
+		externalConfig["ovn-remote"] = val
+	}
+
+	// TODO(ihar) do we need ovn- prefixes for all of the settings?
+	externalConfig["ovn-encap-type"] = instance.Spec.ExternalIDS.OvnEncapType
+
+	// Extract some values from SB DbCluster CRD
+
+	// TODO(ihar) I am not sure these should be aligned (we may want to use different tuning for EDP?)
+	externalConfig["ovn-of-probe-interval"] = strconv.Itoa(int(sbCluster.Spec.InactivityProbe))
```
karelyatin, line 297: ovn-openflow-probe-interval?

controllers/ovncontroller_controller.go
karelyatin: I am still not sure how ansible would be using it. I assume ansible itself wouldn't care about how it get's data i.e from config map or some other sources. May be some operator like dataplane/ansibleee operator would prepare the data in required format for ansible usage?

For prefix part i think it should be fine to keep the prefix as that's what is what expected by target system(ovs config external-ids).

booxter: @karelyatin dataplane-operator will mount the ConfigMap data into edpm-ansible environment, see https://docs.google.com/document/d/1tr5KzgT0cB-nCF36kSmEbzxy9i38oLQyfCsGwgQUiY8/edit

That said, I now figured that I should transform this from direct ConfigMap to data: <file>: format, and then dataplane-operator will mount <file> into a file path. This PR will change direction somewhat to accommodate for that.

templates/ovncontroller/config/ovsdb-config
slagle, line 1: Each top level key under data in the ConfigMap will be its own mounted file into the AnsibleEE pod. This is so that operators can create different file mounts (different config files that might need generating, or a file just for ansible vars, etc).

That being said, if these are meant to be ansible vars, you probably want to nest them under a key which would be used as the file name, so the ConfigMap would look like:

```
data:
  ovn-ansible-vars.yaml:
    ovn-remote: foo
    ...etc.
```

The ansible role (edpm-ovn) would then do a include_vars on that mounted file.

booxter: Here is what the latest version here generates:

```
[ihrachys@ansible-fx2-6 install_yamls]$ oc get configmap -n openstack -o yaml ovncontroller-config                                                                                     
apiVersion: v1
data:
  ovsdb-config: |-
    ovn-remote=tcp:10.217.5.144:6642
    ovn-encap-type=geneve
    ovn-openflow-probe-interval=60000
    ovn-remote-probe-interval=60000
kind: ConfigMap
metadata:
  creationTimestamp: "2023-08-15T20:22:25Z"
  labels:
    ovn-controller.openstack.org/name: ovncontroller
    ovn-controller.openstack.org/namespace: openstack
    ovn-controller.openstack.org/uid: 0b11fd5f-ce5b-46fb-a19e-1b5b2747c4b7
  name: ovncontroller-config
  namespace: openstack
  ownerReferences:
  - apiVersion: ovn.openstack.org/v1beta1
    blockOwnerDeletion: true
    controller: true
    kind: OVNController
    name: ovncontroller
    uid: 0b11fd5f-ce5b-46fb-a19e-1b5b2747c4b7
  resourceVersion: "68301"
  uid: 053d2806-623e-4edc-8404-ea63c3f2c64b
```

And yes, I would like ansibleEE to mount ovsdb-config *file* into the pod. It will then be fed to `ovs-vsctl set` command.

booxter: @slagle Please let me know if the plan above makes sense. It effectively ignores the edpm_* settings for the knobs, so I wonder if it's correct. Maybe I should indeed load as ansible variables and then reconcile with whatever is set as edpm_ in node template?

booxter: (The reason why I don't just pass edpm_ ansible names in the ovsdb-config file is because I believe it's better to keep EDPM flavored names out of service operators interface (which the ConfigMap effectively is - for in case something else but edpm-ansible may want to consume it).

booxter:
This is largely complete, only kuttl tests are missing. I will add them after I have it tested end-to-end with edpm-ansible.

booxter:
This should be ready for review. I left some todos to consider for reviewers but these may be followed up later if needed.

controllers/ovncontroller_controller.go
```diff
@@ -521,6 +551,63 @@ func (r *OVNControllerReconciler) generateServiceConfigMaps(
...
+		{
+			Name:          fmt.Sprintf("%s-config", instance.Name),
+			Namespace:     instance.Namespace,
+			Type:          util.TemplateTypeConfig,
+			InstanceType:  instance.Kind,
+			Labels:        cmLabels,
+			ConfigOptions: externalTemplateParameters,
+		},
+	}
+	return configmap.EnsureConfigMaps(ctx, h, instance, cms, envVars)
```
gibizer, line 587: Is there sensitive data (e.g. passwords) in this config map? In nova the transport_url contain the rabbitmq password so we decided to generate a Secret instead of the ConfigMap.

booxter: No. There is no password authentication for OVN databases. As long as you can get to the port, you can do whatever you want.

gibizer: OK. Then it is a non issue for OVN

tests/functional/ovncontroller_controller_test.go
```diff
@@ -335,6 +352,137 @@ var _ = Describe("OVNController controller", func() {
...
+			Eventually(func() corev1.ConfigMap {
+				return *th.GetConfigMap(externalCM)
+			}, timeout, interval).ShouldNot(BeNil())
+
+			Expect(th.GetConfigMap(externalCM).Data["ovsdb-config"]).Should(
+				ContainSubstring("ovn-remote=%s", externalSBEndpoint))
+			Expect(th.GetConfigMap(externalCM).Data["ovsdb-config"]).Should(
+				ContainSubstring("ovn-encap-type=%s", "geneve"))
+
+			// TODO(ihar) shouldn't these somehow get "default" values from kubebuilder?
```
karelyatin, line 384: we need to fix CreateOVNDBClusters to use CreateUnstructured to get it work, /me will send the patch

templates/ovncontroller/config/ovsdb-config
```diff
@@ -0,0 +1,4 @@
+ovn-remote={{ .OvnRemote }}
+ovn-encap-type={{ .OvnEncapType }}
```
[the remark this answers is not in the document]
booxter: https://github.com/openstack-k8s-operators/edpm-ansible/pull/288

tests/functional/ovncontroller_controller_test.go
booxter: I assume this is https://github.com/openstack-k8s-operators/ovn-operator/pull/112

If it lands before this PR, I will rebase.

controllers/ovncontroller_controller.go
```diff
@@ -432,9 +434,37 @@ func (r *OVNControllerReconciler) reconcileNormal(ctx context.Context, instance
...
 	// create DaemonSet - end
 
+	sbCluster, err := v1beta1.GetDBClusterByType(ctx, helper, instance.Namespace, map[string]string{}, v1beta1.SBDBType)
+	if err != nil {
+		r.Log.Info("No SB OVNDBCluster defined, deleting external ConfigMap")
+		cleanupConfigMapErr := r.deleteExternalConfigMaps(ctx, helper, instance)
+		if cleanupConfigMapErr != nil {
+			r.Log.Error(cleanupConfigMapErr, "Failed to delete external ConfigMap")
+		}
+		return ctrl.Result{}, err
```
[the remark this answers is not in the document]
booxter: Good point, and I think I should return cleanupConfigMapErr if it happened.

booxter:
Changes:
- remove timeout values from configmap
- change format to yaml, edpm will load it as facts

karelyatin (approved):

PR #110, opened 2023-08-19
switch to structured logging

pinikomarov:
This automatically adds additional fields like reconcile_id etc.. from the controller context.

before :

`2023-05-18T01:53:14+03:00 INFO  controllers.KeystoneAPI Reconciled Service init successfully`

after:

`2023-10-19T10:41:34.351+0300    INFO    Controllers.OVNController       Reconciling Service upgrade     {"controller": "ovncontroller", "controllerGroup": "ovn.openstack.org", "controllerKind": "OVNController", "OVNController": {"name":"ovncontroller","namespace":"openstack"}, "namespace": "openstack", "name": "ovncontroller", "reconcileID": "b8e9495c-14f3-461d-ad0e-0e03f193bf4f"}
`

*by using per reconcile function respective logger objects, the intention is to lay the ground for parallel reconciliation in future and avoid race conditions as ctx objects are reconcile run specific.

### DIFF
--- a/controllers/ovncontroller_controller.go
+++ b/controllers/ovncontroller_controller.go
@@ -32,6 +32,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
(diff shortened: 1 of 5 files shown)


karelyatin (commented):
/ok-to-test

slawqo (approved):

PR #111, opened 2023-08-22
[local webhook] Adjust firewall only if binary exists

karelyatin:
The script currently fails on systems where
firewall-cmd binary do not exist. Let's set the
firewall rules only when binary exists.


abays:
While I agree that the error due to the lack of the binary is a problem, I guess we'll assume that the user either has to manually modify the firewall itself or that perhaps no change to it is even needed?

abays (approved):
/lgtm

PR #127, opened 2023-09-26
Add empty spec key for ovncontroller sample

karelyatin:
make ovn_deploy fails since [1] as oc kustomize
fails with missing path.
This patch adds empty spec key to get it working,
alternative would be to update install_yamls to
create kustomization.yaml specific to ovn.

[1] https://github.com/openstack-k8s-operators/ovn-operator/pull/64


slawqo (approved):

booxter:
Thank you for taking care of the breakage.

PR #132, opened 2023-10-02
Remote routev1 dependency

gibizer:
Now that the Route creation is moved to openstack-operator the ovn-operator does not depend on the routev1 module.


abays (approved):
/lgtm

PR #148, opened 2023-10-30
northd: run as openvswitch:openvswitch

booxter:
The container is now unprivileged. It also no longer uses kolla startup path and no longer relies on configMap to pass configuration (instead, command line args are used).


booxter:
I'm also removing "Hash" field here because I don't think it's needed now that northd doesn't consume ConfigMaps. But please reviewers, take notice and confirm.

booxter:
I'll be splitting https://github.com/openstack-k8s-operators/ovn-operator/pull/76 into pieces in several steps. It's the first one and the least controversial I believe.

slawqo (approved):

PR #153, opened 2023-11-10
Create External CM only when DBClusters have NetworkAttachment

karelyatin:
DBAddress is only set when DBClusters are created with a NetworkAttachment. Same condition is now honored while create External Config Map.

This also fixes the case where ovn-configuration job was not created when DBClusters have no NetworkAttachment.

### DIFF
--- a/api/v1beta1/ovndbcluster_types.go
+++ b/api/v1beta1/ovndbcluster_types.go
@@ -197,7 +197,7 @@ func (instance OVNDBCluster) GetInternalEndpoint() (string, error) {
 }
 
 func (instance OVNDBCluster) GetExternalEndpoint() (string, error) {
-	if instance.Status.DBAddress == "" {
+	if instance.Spec.NetworkAttachment != "" && instance.Status.DBAddress == "" {
 		return "", fmt.Errorf("external DBEndpoint not ready yet for %s", instance.Spec.DBType)
 	}
 	return instance.Status.DBAddress, nil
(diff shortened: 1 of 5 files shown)


api/v1beta1/ovndbcluster_types.go
```diff
@@ -197,7 +197,7 @@ func (instance OVNDBCluster) GetInternalEndpoint() (string, error) {
 }
 
 func (instance OVNDBCluster) GetExternalEndpoint() (string, error) {
-	if instance.Status.DBAddress == "" {
+	if instance.Spec.NetworkAttachment != "" && instance.Status.DBAddress == "" {
 		return "", fmt.Errorf("external DBEndpoint not ready yet for %s", instance.Spec.DBType)
```
booxter, line 201: is it not ready *yet*, or will it not be ever available (with the spec without attachment)? Should these two branches then have separate error messages? What do you think?

booxter (changes requested):

api/v1beta1/ovndbcluster_types.go
karelyatin: I think when DB Cluster have No NetworkAttachment we don't need to raise any kind of error as with it DBAddress will always be unset.

booxter: If you don't return an error at all, then `deleteExternalConfigMaps` in line 461 will never be called to clean up obsolete external CM. Is it what you intend here?

booxter (changes requested):
I'd like to understand if we shouldn't clean up configMap on detachment from NAD.

karelyatin:
> I'd like to understand if we shouldn't clean up configMap on detachment from NAD.

Update the PR to delete External CM when NAD is detached from SB DB along with test.

slawqo (approved):
/lgtm

PR #164, opened 2023-11-22
Include status.networkAttachments in PrinterColumns

karelyatin:
Printing the status.networkAttachments instead of the spec part is more useful as it dumps pod IPs.


luis5tb:
nice catch!

luis5tb:
lgtm

slawqo (approved):

PR #175, opened 2023-12-08
Run ovn ovsdb-servers as anyuid

booxter:
We may explore tightening it further to restricted-v2 later.

The change gets rid of kolla in startup path in line with:

https://github.com/openstack-k8s-operators/docs/pull/77


slawqo (approved):

booxter:
CI is temporarily down due to RDO Quay upgrade. I'll recheck later.

PR #176, opened 2023-12-08
Run northd as scc:restricted-v2

booxter:
It's a regular tcp client and has no special requirements as to permissions.


slawqo (approved):

karelyatin (commented):
/lgtm

PR #180, opened 2023-12-14
Tighten ovn dbclusters to scc=restricted-v2

booxter:
The only obstacle was the logfile.

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -222,14 +222,6 @@ rules:
   - security.openshift.io
   resourceNames:
   - anyuid
-  resources:
-  - securitycontextconstraints
-  verbs:
-  - use
-- apiGroups:
-  - security.openshift.io
-  resourceNames:
-  - anyuid
   - privileged
   resources:
   - securitycontextconstraints
--- a/controllers/ovndbcluster_controller.go
+++ b/controllers/ovndbcluster_controller.go
@@ -91,7 +91,7 @@ func (r *OVNDBClusterReconciler) GetLogger(ctx context.Context) logr.Logger {
 // +kubebuilder:rbac:groups="rbac.authorization.k8s.io",resources=roles,verbs=get;list;watch;create;update
 // +kubebuilder:rbac:groups="rbac.authorization.k8s.io",resources=rolebindings,verbs=get;list;watch;create;update
 // service account permissions that are needed to grant permission to the above
-// +kubebuilder:rbac:groups="security.openshift.io",resourceNames=anyuid,resources=securitycontextconstraints,verbs=use
+// +kubebuilder:rbac:groups="security.openshift.io",resourceNames=restricted-v2,resources=securitycontextconstraints,verbs=use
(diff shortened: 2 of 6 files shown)


karelyatin:
With this i see couple of operation not permitted messages which were not there before https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openstack-k8s-operators_ovn-operator/180/pull-ci-openstack-k8s-operators-ovn-operator-main-ovn-operator-build-deploy-kuttl/1742966878868869120/artifacts/ovn-operator-build-deploy-kuttl/openstack-k8s-operators-gather/artifacts/must-gather/quay-io-openstack-k8s-operators-openstack-must-gather-sha256-ddf759222446dc652c0d70ace352eac321c9a7fc990c78254257026f39e535c4/namespaces/openstack/pods/ovsdbserver-nb-0/logs/ovsdbserver-nb.log , can you check if could get rid of those.

booxter:
@karelyatin This may be problematic.

1. I am able to remove `install: cannot change owner and permissions of '/var/log/ovn': No such file or directory` by setting `OVN_LOGDIR` to `/tmp` through `envVars`. (Will include it in the next iteration.)

2. For `chown: changing ownership of '/tmp': Operation not permitted`, this happens because `ovn-ctl` attempts to change the ownership of a number of files as follows:

```
    # Set the owner of the ovn_dbdir (with -R option) to OVN_USER if set.
    # This is required because the ovndbs are created with root permission
    # if not present when create_cluster/upgrade_db is called.
    INSTALL_USER="root"
    INSTALL_GROUP="root"
    [ "$OVN_USER" != "" ] && INSTALL_USER="${OVN_USER%:*}"
    [ "${OVN_USER##*:}" != "" ] && INSTALL_GROUP="${OVN_USER##*:}"

    chown -R $INSTALL_USER:$INSTALL_GROUP $ovn_dbdir
    chown -R $INSTALL_USER:$INSTALL_GROUP $OVN_RUNDIR
    chown -R $INSTALL_USER:$INSTALL_GROUP $ovn_logdir
    chown -R $INSTALL_USER:$INSTALL_GROUP $ovn_etcdir
```

But if I pass `--ovn-user=$(id -u):$(id -g)` to `ovn-ctl`, then it fails to start `ovsdb-server` because of:

```
ovsdb-server: /tmp/ovnnb_db.pid: only root can use --user option
```

```
    uid = getuid();

    if (geteuid() || uid) {
        VLOG_FATAL("%s: only root can use --user option", pidfile);
    }
```

Since we are not root (the goal of the PR is to restrict the pod), this seems impossible.

It seems that `ovn-ctl` (and `ovs-ctl`) implies that it's being run as `root`, and while it works, the permissions on the files cannot be changed. (Which in my mind is expected.)

booxter:
@karelyatin I updated to remove the error message about the missing /var/log/ovn directory; but for the rest, I don't know if we can do much except patching ovn-ctl itself to gracefully handle non-root execution.

karelyatin:
Thanks Ihar for digging it further and adding a Jira tracker

PR #182, opened 2023-12-15
Clean up volumes list for ovn-controller pod

booxter:
These seem unnecessary; and they block our progress on moving from HostPaths to e.g. PVCs, which would allow us to disable privileged mode (/home hostPaths seems to carry selinux labels that are not available for the containers.)


booxter:
var-run is consumed by config-job for nicMappings etc. Need to reconcile with it. Maybe we don't need a config-job and can squash it into the same statefulset somehow?..

booxter:
Rebased on top of https://github.com/openstack-k8s-operators/ovn-operator/pull/195

booxter:
Some of the volumes can be cleaned up with no bad effect. But at least the volume that passes over the ovsdb-server AF_UNIX socket to configJob via a file cannot be removed until we switch this communication channel to AF_INET sockets. This is technically possible but requires significant work to e.g. deploy SSL certificates for the channel.

This PR may still be useful, but I should drop the removal of the socket volume from it first before we can proceed with a partial volume list cleanup.

booxter:
I'm not going to pursue this. Feel free to fork if needed.

PR #187, opened 2023-12-19
standardize Makefile operator-sdk

yazug:
adjust location of several Makefile bits
sync Makefile from skeleton Makefile made with operator-sdk init

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,6 +46,10 @@ ifeq ($(USE_IMAGE_DIGESTS), true)
 	BUNDLE_GEN_FLAGS += --use-image-digests
 endif
 
+# Set the Operator SDK version to use. By default, what is installed on the system is used.
+# This is useful for CI or a project to utilize a specific version of the operator-sdk toolkit.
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -224,8 +229,25 @@ ginkgo: $(GINKGO) ## Download ginkgo locally if necessary.
 $(GINKGO): $(LOCALBIN)
 	test -s $(LOCALBIN)/ginkgo || GOBIN=$(LOCALBIN) go install github.com/onsi/ginkgo/v2/ginkgo
 
+.PHONY: operator-sdk
+OPERATOR_SDK ?= $(LOCALBIN)/operator-sdk
+operator-sdk: ## Download operator-sdk locally if necessary.
```
karelyatin, line 234: It needs to be dropped from L338

karelyatin (changes requested):

Makefile
yazug: good catch, dropped the duplicated section.  This is the reason for standardizing where things are better.

karelyatin (approved):
/lgtm

PR #213, opened 2024-01-30
Handle the physical network changes "ovn-bridge-mappings"

ralonsoh:
Now when the physical network changes, the list of OVS bridges is compared to the current one:
* The new bridges are added to OVS.
* The old bridges are deleted along with the patch port to "br-int".
* The new "ovn-bridge-mappings" is set in the local OVS Open vSwitch register.

Closes: OSPRH-312


templates/ovncontroller/bin/functions
```diff
@@ -52,31 +52,59 @@ function configure_external_ids {
...
+
+    # Delete the old bridges not longer present in "OvnBridgeMappings" and the
+    # patch ports in "br-int".
+    for br_name in ${br_to_add}; do
+        # List all ports "type=patch", retrieve the "options:peer" parameter
+        # and delete the peer port. The "$port" will be deleted during the
+        # bridge deletion.
+        #   $ ovs-vsctl --column=options --bare find interface name=patch-provnet1-to-br-int type=patch
+        #     peer=patch-br-int-to-provnet1
+        for port in $(ovs-vsctl list-port ${br-name}); do
```
karelyatin, line 97: list-ports?
${br-name} --> ${br_name}

templates/ovncontroller/bin/functions
```diff
@@ -52,31 +52,59 @@ function configure_external_ids {
...
+
+    # Current configured bridges.
+    ovn_bms=$(ovs-vsctl --if-exists get open . external_ids:ovn-bridge-mappings|tr -d '"')
+    local br_current=""
+    for bm in ${ovn_bms/,/ }; do
+        if [ -z "$br_current"]; then
+            br_current=${bm##*:}
+        else
+            br_current="${br_current} ${bm##*:}"
         fi
```
karelyatin, line 76: seems it no longer set or remove external_ids:ovn-bridge-mappings

karelyatin (commented):
pre commit failures are related:-

templates/ovncontroller/bin/functions:80:1: E042 local declaration hides errors
templates/ovncontroller/bin/functions:81:1: E042 local declaration hides errors
2 bashate error(s) found

https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openstack-k8s-operators_ovn-operator/213/pull-ci-openstack-k8s-operators-ovn-operator-main-precommit-check/1752585774924042240/build-log.txt

templates/ovncontroller/bin/functions
```diff
@@ -52,31 +52,59 @@ function configure_external_ids {
...
+
+    # Add the new bridges.
+    for br_name in ${br_to_add}; do
+        ovs-vsctl --may-exist add-br ${br_name}
+        ovs-vsctl --may-exist add-port ${br_name} ${br_name##*-}
+    done
+
+    # Delete the old bridges not longer present in "OvnBridgeMappings" and the
+    # patch ports in "br-int".
+    for br_name in ${br_to_add}; do
```
slawqo, line 91: shouldn't You iterate here over ${br_to_delete} list?

templates/ovncontroller/bin/functions
```diff
@@ -52,31 +52,59 @@ function configure_external_ids {
...
+    local br_current=""
+    for bm in ${ovn_bms/,/ }; do
+        if [ -z "$br_current"]; then
+            br_current=${bm##*:}
+        else
+            br_current="${br_current} ${bm##*:}"
         fi
     done
-    if [ -n "$OvnBridgeMappings" ]; then
-        ovs-vsctl set open . external-ids:ovn-bridge-mappings=${OvnBridgeMappings}
```
slawqo, line 67: Yatin is correct, I think You shouldn't remove that

templates/ovncontroller/bin/functions
ralonsoh: Exactly!

templates/ovncontroller/bin/functions
ralonsoh: Right!

templates/ovncontroller/bin/functions
ralonsoh: Yes, I missed that part.

templates/ovncontroller/bin/functions
ralonsoh: Right

templates/ovncontroller/bin/functions
```diff
@@ -52,31 +52,59 @@ function configure_external_ids {
...
+        # List all ports "type=patch", retrieve the "options:peer" parameter
+        # and delete the peer port. The "$port" will be deleted during the
+        # bridge deletion.
+        #   $ ovs-vsctl --column=options --bare find interface name=patch-provnet1-to-br-int type=patch
+        #     peer=patch-br-int-to-provnet1
+        for port in $(ovs-vsctl list-port ${br-name}); do
+            options=$(ovs-vsctl --column=options --bare find interface ${port} type=patch)
+            if [ -n $options ]; then
+                for option in ${options}; do
+                    if [[ "$option" =~ .*"peer".* ]]; then
```
booxter, line 101: AFAIU this is unstrict match. Please match exactly with "peer" string (no substring matches).

templates/ovncontroller/bin/functions
```diff
@@ -52,31 +52,59 @@ function configure_external_ids {
...
+    # patch ports in "br-int".
+    for br_name in ${br_to_add}; do
+        # List all ports "type=patch", retrieve the "options:peer" parameter
+        # and delete the peer port. The "$port" will be deleted during the
+        # bridge deletion.
+        #   $ ovs-vsctl --column=options --bare find interface name=patch-provnet1-to-br-int type=patch
+        #     peer=patch-br-int-to-provnet1
+        for port in $(ovs-vsctl list-port ${br-name}); do
+            options=$(ovs-vsctl --column=options --bare find interface ${port} type=patch)
+            if [ -n $options ]; then
```
booxter, line 99: please add "" around $options, in case the result of the command execution is empty

booxter: that said, this line is probably redundant because match against "peer" on an empty string will return false; rendiring the for-loop a no-op.

templates/ovncontroller/bin/functions
```diff
@@ -52,31 +52,59 @@ function configure_external_ids {
...
+                for option in ${options}; do
+                    if [[ "$option" =~ .*"peer".* ]]; then
+                        ovs-vsctl --if-exists del-port br-int $port ${br_name##*=}
```
booxter, line 102: I'd think you'd delete the peer port from the option here, but AFAIU you instead delete the $port itself?

booxter: Please, for those of us who struggle with shell syntax, would you mind explaining in a comment above what ##*= is supposed to do? Thanks.

templates/ovncontroller/bin/functions
```diff
@@ -52,31 +52,59 @@ function configure_external_ids {
...
-            ovs-vsctl --if-exists del-br ${bm##*:}
-        done
-        ovs-vsctl --if-exists remove open . external_ids ovn-bridge-mappings
 
-        # If NicMappings not defined or cleared, let's ensure patch ports are removed
-        patch_ports=$(ovs-vsctl --columns=name --bare find interface type=patch)
-        for port in ${patch_ports}; do
-            ovs-vsctl --if-exists del-port br-int $port
+    # Bridges to add and delete.
+    local br_to_delete="$(comm -23 <(sort <(echo $br_current | xargs -n1)) <(sort <(echo $br_new | xargs -n1)))"
```
booxter, line 80: these lines are duplicative and not easy to parse. Could you please move the common parts into a function with a comprehensible name? E.g. you could have a function called `get_unique` that receives two lists and returns the difference; then a function called `sort_list` that receives a string and returns an array of strings (sorted), then you could:

```
local br_to_delete=$(get_unique $(sort_list $br_current $br_new))
local br_to_add=$(get_unique $(sort_list $br_new $br_current))
```

booxter (changes requested):
I am concerned that we don't have test coverage for the script, so whatever the issues we introduce here, they won't be caught by CI. I wonder if we can do anything about it (not sure if kuttl could be bent to cover the scenario...)

karelyatin:
> I am concerned that we don't have test coverage for the script, so whatever the issues we introduce here, they won't be caught by CI. I wonder if we can do anything about it (not sure if kuttl could be bent to cover the scenario...)

I think yes could do some kuttl tests here. Recently with other PR https://github.com/openstack-k8s-operators/ovn-operator/pull/237 i added test for similar scenario.

slawqo:
recheck

templates/ovncontroller/bin/functions
```diff
@@ -49,34 +49,73 @@ function configure_external_ids {
...
-        # If NicMappings not defined or cleared, let's ensure patch ports are removed
-        patch_ports=$(ovs-vsctl --columns=name --bare find interface type=patch)
-        for port in ${patch_ports}; do
-            ovs-vsctl --if-exists del-port br-int $port
+    # Current configured bridges.
+    ovn_bms=$(ovs-vsctl --if-exists get open . external_ids:ovn-bridge-mappings|tr -d '"')
+    local br_current=""
+    for bm in ${ovn_bms/,/ }; do
+        if [ -z "$br_current"]; then
+            br_current=${bm##*:}
```
booxter, line 84: better to extract ##*: once before the `if`, then reuse

templates/ovncontroller/bin/functions
```diff
@@ -49,34 +49,73 @@ function configure_external_ids {
...
     local OvnBridgeMappings=""
+    local br_new=""
     for physicalNetwork in ${PhysicalNetworks}; do
         br_name="br-${physicalNetwork}"
-        ovs-vsctl --may-exist add-br ${br_name}
-        ovs-vsctl --may-exist add-port ${br_name} ${physicalNetwork}
         bridgeMapping="${physicalNetwork}:${br_name}"
         if [ -z "$OvnBridgeMappings"]; then
             OvnBridgeMappings=$bridgeMapping
+            br_new=$br_name
```
booxter, line 66: (no action required) we could have a join() function that would receive an array and join the results. like:

```
joinWith() {
  local IFS="$1"
  shift
  echo "$*"
}

joinWith , bridgeMappings
joinWith ' ' bridges
```

we could have that used below to calculate br_current too...

templates/ovncontroller/bin/functions
```diff
@@ -49,34 +49,73 @@ function configure_external_ids {
     fi
 }
 
+# Returns the set difference between $1 and $2
+function set_difference {
```
booxter, line 53: nice!

booxter:
I think this is good in general. I'd try to make even more helper functions with common code and simplify the "main" entrypoint function to the level where most of the actual work would be done in helpers, but that's not required. (The function is already doing a lot of work anyway.)

I think several things are still missing:
- precheck fails on the change, complaining about variables overloaded? please check
- kuttl job seems to be unhappy about the old state of this PR; could you please rebase to latest?
- I think we should add some kuttl scenarios to prove this change works as expected.

ralonsoh:
Superseded by https://github.com/openstack-k8s-operators/ovn-operator/pull/248

PR #215, opened 2024-01-31
Fix default handling for EnableChassisAsGateway

karelyatin:
When OVNController CR is created with empty spec or empty spec.external-ids EnableChassisAsGateway is set to "false" even we have "true" set as kubebuilder default for it.

[1] provide more details on this behavior as kubebuilder default is different here then the golang default for the bool type. Handle it by changing EnableChassisAsGateway to bool pointer type.

Other alternative could be to duplicate default at spec level like:- // +kubebuilder:default={ovn-bridge: br-int}
ExternalIDS OVSExternalIDs `json:"external-ids"`

[1] https://github.com/gibizer/operator-lint/tree/main/linters/crd/C003


api/v1beta1/ovncontroller_types.go
```diff
@@ -158,7 +158,7 @@ type OVSExternalIDs struct {
 
 	// +kubebuilder:validation:Optional
 	// +kubebuilder:default=true
-	EnableChassisAsGateway bool `json:"enable-chassis-as-gateway"`
+	EnableChassisAsGateway *bool `json:"enable-chassis-as-gateway,omitempty"`
```
gibizer, line 161: I agree that it needs to be a pointer as you want to make it optional. I'm not sure you need the omitempty after you made it to a pointer.

karelyatin: thx dropped it, as seems not needed.

abays (approved):
/lgtm

PR #217, opened 2024-02-02
Add type label into headless/normal services

averdagu:
Adding type label to easier search between headless or normal service and use them to filter only the normal services during scale-down service removal.

Resolves: [OSPRH-4120](https://issues.redhat.com//browse/OSPRH-4120)

### DIFF
--- a/controllers/ovndbcluster_controller.go
+++ b/controllers/ovndbcluster_controller.go
@@ -533,9 +533,10 @@ func (r *OVNDBClusterReconciler) reconcileServices(
 	//
 	// Ensure the ovndbcluster headless service Exists
 	//
+	headlessServiceLabels := util.MergeMaps(serviceLabels, map[string]string{"type": v1beta1.ServiceHeadlessType})
 
 	headlesssvc, err := service.NewService(
-		ovndbcluster.HeadlessService(serviceName, instance, serviceLabels),
+		ovndbcluster.HeadlessService(serviceName, instance, headlessServiceLabels, serviceLabels),
 		time.Duration(5)*time.Second,
 		nil,
 	)
@@ -559,12 +560,13 @@ func (r *OVNDBClusterReconciler) reconcileServices(
 		//
 		// Create the ovndbcluster pod service if none exists
 		//
-		ovndbServiceLabels := map[string]string{
+		ovndbSelectorLabels := map[string]string{
 			common.AppSelector:                   serviceName,
 			"statefulset.kubernetes.io/pod-name": ovnPod.Name,
 		}
+		ovndbServiceLabels := util.MergeMaps(ovndbSelectorLabels, map[string]string{"type": v1beta1.ServiceClusterType})
 		svc, err := service.NewService(
-			ovndbcluster.Service(ovnPod.Name, instance, ovndbServiceLabels),
+			ovndbcluster.Service(ovnPod.Name, instance, ovndbServiceLabels, ovndbSelectorLabels),
 			time.Duration(5)*time.Second,
 			nil,
 		)
@@ -581,11 +583,12 @@ func (r *OVNDBClusterReconciler) reconcileServices(
 	}
 
 	// Delete any extra services left after scale down
+	clusterServiceLabels := util.MergeMaps(serviceLabels, map[string]string{"type": v1beta1.ServiceClusterType})
 	svcList, err := service.GetServicesListWithLabel(
 		ctx,
 		helper,
 		helper.GetBeforeObject().GetNamespace(),
-		serviceLabels,
+		clusterServiceLabels,
 	)
 	if err == nil && len(svcList.Items) > int(*(instance.Spec.Replicas)) {
 		for i := len(svcList.Items) - 1; i >= int(*(instance.Spec.Replicas)); i-- {
--- a/tests/functional/ovndbcluster_controller_test.go
+++ b/tests/functional/ovndbcluster_controller_test.go
@@ -226,6 +226,40 @@ var _ = Describe("OVNDBCluster controller", func() {
 			DeferCleanup(th.DeleteInstance, instance)
 		})
 
+		It("should create services", func() {
+			internalAPINADName := types.NamespacedName{Namespace: namespace, Name: "internalapi"}
+			nad := th.CreateNetworkAttachmentDefinition(internalAPINADName)
+			DeferCleanup(th.DeleteInstance, nad)
+
+			statefulSetName := types.NamespacedName{
+				Namespace: namespace,
+				Name:      "ovsdbserver-sb",
+			}
+			th.SimulateStatefulSetReplicaReadyWithPods(
+				statefulSetName,
+				map[string][]string{namespace + "/internalapi": {"10.0.0.1"}},
+			)
+			// Cluster is created with 1 replica, serviceListWithoutTypeLabel should be 2:
+			// - ovsdbserver-sb   (headless type)
+			// - ovsdbserver-sb-0 (cluster type)
+			Eventually(func(g Gomega) {
+				serviceListWithoutTypeLabel := GetServicesListWithLabel(namespace)
+				g.Expect(len(serviceListWithoutTypeLabel.Items)).To(Equal(2))
+			}).Should(Succeed())
+
+			Eventually(func(g Gomega) {
+				serviceListWithClusterType := GetServicesListWithLabel(namespace, map[string]string{"type": "cluster"})
+				g.Expect(len(serviceListWithClusterType.Items)).To(Equal(1))
+				g.Expect(serviceListWithClusterType.Items[0].Name).To(Equal("ovsdbserver-sb-0"))
+			}).Should(Succeed())
+
+			Eventually(func(g Gomega) {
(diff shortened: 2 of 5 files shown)


controllers/ovndbcluster_controller.go
```diff
@@ -581,11 +586,15 @@ func (r *OVNDBClusterReconciler) reconcileServices(
 	}
 
 	// Delete any extra services left after scale down
+	normalServiceLabels := map[string]string{
+		common.AppSelector: serviceName,
+		"type":             "normal",
```
averdagu, line 591: Not a big fan of the "normal" name, but the other name that I thought was "service" but can be "confusing" since both Headless and "normal" service are both services. I'm open to new namings

lewisdenny:
Hey @averdagu I started looking into this out of curiosity but didn't get time to finish, just sharing my findings: 
[1] 
```
  - lastTransitionTime: "2024-02-02T15:32:37Z"
    message: OpenStackControlPlane Neutron in progress
    reason: Requested
    severity: Info
    status: "False"
    type: Ready
```
[2]
```
Exception: Could not retrieve schema from tcp:ovsdbserver-sb-0.openstack.svc.cluster.local:6642
```

[1] https://logserver.rdoproject.org/17/217/d929c50b05f7ad4137efda0dccbf8b49b1e75d5f/github-check/ovn-operator-tempest-multinode/1e0c3ef/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/crs/openstackcontrolplanes.core.openstack.org/openstack-galera-network-isolation.yaml

[2]https://logserver.rdoproject.org/17/217/d929c50b05f7ad4137efda0dccbf8b49b1e75d5f/github-check/ovn-operator-tempest-multinode/1e0c3ef/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/pods/neutron-7cfb6b45dd-pd95b/neutron-7cfb6b45dd-pd95b-describe

lewisdenny:
recheck

controllers/ovndbcluster_controller.go
```diff
@@ -533,9 +533,16 @@ func (r *OVNDBClusterReconciler) reconcileServices(
 	//
 	// Ensure the ovndbcluster headless service Exists
 	//
+	headlessServiceLabels := map[string]string{
```
karelyatin, line 536: can just extend serviceLabels here instead of duplicating.

controllers/ovndbcluster_controller.go
```diff
@@ -533,9 +533,16 @@ func (r *OVNDBClusterReconciler) reconcileServices(
...
+		common.AppSelector: serviceName,
+		"type":             "headless",
+	}
+	headlessSelectorLabels := map[string]string{
```
karelyatin, line 540: can use serviceLabels instead of defining new variable here.

controllers/ovndbcluster_controller.go
```diff
@@ -562,9 +569,14 @@ func (r *OVNDBClusterReconciler) reconcileServices(
 		ovndbServiceLabels := map[string]string{
 			common.AppSelector:                   serviceName,
```
karelyatin, line 570: same here, can extend ovndbSelectorLabels

controllers/ovndbcluster_controller.go
```diff
@@ -581,11 +593,15 @@ func (r *OVNDBClusterReconciler) reconcileServices(
 	}
 
 	// Delete any extra services left after scale down
+	normalServiceLabels := map[string]string{
```
karelyatin, line 596: normal vs cluster in name. wdyt to use"regular" as type?

averdagu: The problem with "normal" or "regular" is that can be missleading, as it doesn't mean anything inside kubernetes, what's normal or regular? That's why I changed for Cluster (As one of the differences bw headless services and non headless services is that one has Cluster Ip and the other has not). That's why "type: cluster" is the best name I found so far). wdyt?

Also, true, will change normalServiceLabels to clusterServiceLabels, as there's a mismatch on naming :)

controllers/ovndbcluster_controller.go
averdagu: Extending map from serviceLabels would modify servicelabels, we could use maps.Clone https://pkg.go.dev/maps

karelyatin: i was considering like using something ilke https://github.com/openstack-k8s-operators/lib-common/blob/c9467a8fbbfc3e244207cd3d435975aca298d33d/modules/common/util/map.go#L86

controllers/ovndbcluster_controller.go
```diff
@@ -533,9 +533,10 @@ func (r *OVNDBClusterReconciler) reconcileServices(
 	//
 	// Ensure the ovndbcluster headless service Exists
 	//
+	headlessServiceLabels := util.MergeMaps(serviceLabels, map[string]string{"type": "cluster"})
```
karelyatin, line 536: type: headless

controllers/ovndbcluster_controller.go
```diff
@@ -581,11 +583,15 @@ func (r *OVNDBClusterReconciler) reconcileServices(
 	}
 
 	// Delete any extra services left after scale down
+	clusterServiceLabels := map[string]string{
```
karelyatin, line 586: clusterServiceLabels := util.MergeMaps(serviceLabels, map[string]string{"type": "cluster"})

karelyatin (changes requested):

controllers/ovndbcluster_controller.go
booxter: this should probably be a const in v1betav1/ module

booxter (changes requested):
My main request here is that a test case should be added to demonstrate the failing scenario.

(It would also be nice to have some documentation as to which services are DNS records are created by the operator, because it becomes hard to follow the changes in the area. But this is not a requirement for this PR, just an observation that some docs are probably due and I would appreciate if you could propose some for this repo.)

controllers/ovndbcluster_controller.go
averdagu: Thanks Yatin and Ihar, will apply changes

controllers/ovndbcluster_controller.go
averdagu: Done

tests/functional/ovndbcluster_controller_test.go
```diff
@@ -226,6 +227,51 @@ var _ = Describe("OVNDBCluster controller", func() {
...
+				statefulSetName,
+				map[string][]string{namespace + "/internalapi": {"10.0.0.1"}},
+			)
+			service := types.NamespacedName{
+				Namespace: OVNDBClusterName.Namespace,
+				Name:      "ovsdbserver-sb",
+			}
+			// Even with the Eventually OVN takes to long to reconcile and create both services
+			// with this extra 2 seconds by the time the check is done, services will be already created
+			time.Sleep(2 * time.Second)
```
karelyatin, line 249: this doesn't seem needed, Can you try dropping this workaround + fixing operator-lint issues

tests/functional/ovndbcluster_controller_test.go
```diff
@@ -226,6 +227,51 @@ var _ = Describe("OVNDBCluster controller", func() {
...
+
+			statefulSetName := types.NamespacedName{
+				Namespace: namespace,
+				Name:      "ovsdbserver-sb",
+			}
+			th.SimulateStatefulSetReplicaReadyWithPods(
+				statefulSetName,
+				map[string][]string{namespace + "/internalapi": {"10.0.0.1"}},
+			)
+			service := types.NamespacedName{
```
karelyatin, line 243: namespacename object doesn't look necessary, can just pass the namespace as that's what used by GetServicesListWithLabel.

tests/functional/ovndbcluster_controller_test.go
```diff
@@ -226,6 +227,51 @@ var _ = Describe("OVNDBCluster controller", func() {
...
+			// Cluster is created with 1 replica, serviceListWithoutTypeLabel should be 2:
+			// - ovsdbserver-sb   (headless type)
+			// - ovsdbserver-sb-0 (cluster type)
+			Eventually(func(g Gomega) {
+				serviceListWithoutTypeLabel := GetServicesListWithLabel(service)
+				Expect(len(serviceListWithoutTypeLabel.Items)).To(Equal(2))
+			}).Should(Succeed())
+
+			Eventually(func(g Gomega) {
+				clusterService := types.NamespacedName{
```
karelyatin, line 259: same

tests/functional/ovndbcluster_controller_test.go
averdagu: Done

tests/functional/ovndbcluster_controller_test.go
averdagu: Done

tests/functional/ovndbcluster_controller_test.go
averdagu: Done

averdagu:
recheck

karelyatin (approved):

PR #232, opened 2024-02-20
[ovn-northd] Don't set DeployReady condition with replicas 0

karelyatin:
When OVNNorthd CR is created with 0 replicas, deployment
get's created and DeploymentReady condition was not set which
leads to Northd CR not getting to Ready State.

With this patch, when northd is deployed with 0 replicas
we drop DeploymentReady condition and let Ready condition
get's set for it.

Related-Issue: [OSPRH-2437](https://issues.redhat.com//browse/OSPRH-2437)


karelyatin:
recheck infra issue

slawqo (approved):

PR #233, opened 2024-02-22
Split ovn-controller in separate daemonsets

fernandoroyosanchez:
Split the services ovsdb-server and ovs-vswitchd from current ovn-controller. As this way we would have 2 different
daemonsets, one for the ovn-controller and another one for the ovsdbserver and ovsvswitchd, allowing the update of ovn-controller in a separate way.

Closes: [OSPRH-1932](https://issues.redhat.com//browse/OSPRH-1932)


fernandoroyosanchez:
recheck

fernandoroyosanchez:
recheck

fernandoroyosanchez:
recheck

api/v1beta1/ovncontroller_types.go
```diff
@@ -94,6 +94,9 @@ type OVNControllerStatus struct {
 	// NumberReady of the OVNController instances
 	NumberReady int32 `json:"numberReady,omitempty"`
 
+	// OVS-NumberReady of the OVNController instances
+	OVSNumberReady int32 `json:"ovsNumberReady,omitempty"`
```
slawqo, line 98: if this is for "OVNController instances" why don't You name it like "OVNControllerNumberReady" or something like that? Or should the comment be different and say about ovs-vswitchd + ovsdb-server PODs instances?

controllers/ovncontroller_controller.go
```diff
@@ -553,6 +553,37 @@ func (r *OVNControllerReconciler) reconcileNormal(ctx context.Context, instance
 	instance.Status.DesiredNumberScheduled = dset.GetDaemonSet().Status.DesiredNumberScheduled
 	instance.Status.NumberReady = dset.GetDaemonSet().Status.NumberReady
 
+	// Define a new DaemonSet object for the OVS (ovsdbserver+ovsvswitchd)
+	ovsDaemonSet, err := ovncontroller.CreateOVSDaemonSet(instance, inputHash, serviceLabels, serviceAnnotations)
+	if err != nil {
+		Log.Error(err, "Failed to create OVS DaemonSet")
+		return ctrl.Result{}, err
+	}
+	dset = daemonset.NewDaemonSet(
```
slawqo, line 562: maybe You could use some different variable here, like e.g. "ovsdset" or something like that, wdyt?

controllers/ovncontroller_controller.go
```diff
@@ -611,7 +642,7 @@ func (r *OVNControllerReconciler) reconcileNormal(ctx context.Context, instance
 	}
 
 	// create OVN Config Job - start
-	if instance.Status.NumberReady == instance.Status.DesiredNumberScheduled {
+	if instance.Status.NumberReady == instance.Status.DesiredNumberScheduled && instance.Status.OVSNumberReady == instance.Status.DesiredNumberScheduled {
```
slawqo, line 645: just a question - do we need to wait for both "NumberReady" and "OVSNumberReady" to be as desired? If I'm not mistaken technically it should be enough if we will have OVS PODs ready to run config job and set config knobs in the ovs database, right?

controllers/ovncontroller_controller.go
```diff
@@ -659,7 +690,7 @@ func (r *OVNControllerReconciler) reconcileNormal(ctx context.Context, instance
 		}
 		instance.Status.Conditions.MarkTrue(condition.ServiceConfigReadyCondition, condition.ServiceConfigReadyMessage)
 	} else {
-		Log.Info("OVNController DaemonSet not ready yet. Configuration job cannot be started.")
+		Log.Info("OVNController and OVS DaemonSets not ready yet. Configuration job cannot be started.")
```
slawqo, line 693: nit: technically speaking You should say here "OVNController OR OVS DaemonSet not ready yet"

controllers/ovncontroller_controller.go
fernandoroyosanchez: Of course, it will improve the readability of the code and there is no need to save variables.

controllers/ovncontroller_controller.go
fernandoroyosanchez: yeah, make sense. ConfigJob will run init.sh that basically write config on OVS DB.

api/v1beta1/ovncontroller_types.go
fernandoroyosanchez: I think changing the comment should be enough, copy&paste played a dirty trick here

pkg/ovncontroller/volumes.go
```diff
@@ -2,6 +2,7 @@ package ovncontroller
 
 import (
 	"fmt"
+
```
mtomaska, line 5: what is this extra line for?

fernandoroyosanchez: That has been the linter from the code editor, but I still think it is correct that separation for standard, current package, and vendor imports should be separated by a blank line.

tests/kuttl/common/assert_sample_deployment.yaml
```diff
@@ -292,8 +313,8 @@ commands:
         num_nodes=$(oc get nodes -l node-role.kubernetes.io/worker -o name | wc -l)
         # check the number of ovn-controller pods that are running
         num_pods=$(oc get pods -n $NAMESPACE -l service=ovn-controller -o name --field-selector=status.phase=Running | wc -l)
-        if [ "$num_nodes" != "$num_pods" ]; then
-          echo "Cluster has $num_nodes nodes but OVNController spawned $num_pods pods"
+        if [ "$num_nodes" -ne "$((num_pods/2))" ]; then
```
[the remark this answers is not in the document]
booxter: so what if the number of pods is 5 when there are 2 nodes? your check will pass because / means integer division; but that would be wrong, no? I think you should instead multiply num_nodes by 2 and check against that.

booxter (changes requested):
Directionally it's correct, but there are a number of corners here that should not have been cut

- service labels should be updated to distinguish ovs and ovnc pods
- daemonset.go should be cleaned up to have separate logic for ovs and ovnc
- volumes should be cleaned up to not mount what's not needed

Some more comments included too.

booxter:
I pushed a bunch of patches that I think make this PR more readable and maintainable. I may continue tomorrow.

pkg/ovncontroller/daemonset.go
```diff
@@ -27,30 +26,19 @@ import (
 	"k8s.io/utils/ptr"
 )
 
-// DaemonSet func
-func DaemonSet(
+func CreateOVNDaemonSet(
```
booxter, line 29: these should be renamed because they don't create anything, just prepare an object.

controllers/ovncontroller_controller.go
```diff
@@ -553,6 +553,37 @@ func (r *OVNControllerReconciler) reconcileNormal(ctx context.Context, instance
...
 
+	// Define a new DaemonSet object for the OVS (ovsdbserver+ovsvswitchd)
+	ovsDaemonSet, err := ovncontroller.CreateOVSDaemonSet(instance, inputHash, serviceLabels, serviceAnnotations)
+	if err != nil {
+		Log.Error(err, "Failed to create OVS DaemonSet")
+		return ctrl.Result{}, err
+	}
+	ovsdset := daemonset.NewDaemonSet(
+		ovsDaemonSet,
+		time.Duration(5)*time.Second,
```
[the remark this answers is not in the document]
booxter: Tried that and was not happy about the result which made it less readable. Will keep it as is.

booxter:
@karelyatin @slawqo I am happy enough with this version of the PR. Please take a look. (I haven't dived into func test changes here but they seem ok. But please check these too.)

karelyatin (commented):
/lgtm

karelyatin (approved):

PR #234, opened 2024-02-22
OVN TLS support

olliewalsh:
Add TLS support for ovndbcluster/northd/controller.

Move the service name consts to the API as openstack-operator needs this to create the certs.

Jira: OSPRH-2191


pkg/ovncontroller/daemonset.go
```diff
@@ -32,6 +36,35 @@ func DaemonSet(
 	runAsUser := int64(0)
 	privileged := true
 
+	volumes := GetVolumes(instance.Name, instance.Namespace)
+	commonVolumeMounts := []corev1.VolumeMount{}
+
+	// add CA bundle if defined
+	if instance.Spec.TLS.CaBundleSecretName != "" {
+		volumes = append(volumes, instance.Spec.TLS.CreateVolume())
+		commonVolumeMounts = append(commonVolumeMounts, instance.Spec.TLS.CreateVolumeMounts(nil)...)
```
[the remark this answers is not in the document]
olliewalsh: Full bundle. OVN CA cert is taken from the service secret

controllers/ovncontroller_controller.go
```diff
@@ -199,21 +207,82 @@ func (r *OVNControllerReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
+			builder.WithPredicates(predicate.ResourceVersionChangedPredicate{}),
+		).
 		Complete(r)
 }
 
-func (r *OVNControllerReconciler) reconcileDelete(ctx context.Context, instance *v1beta1.OVNController, helper *helper.Helper) (ctrl.Result, error) {
+func (r *OVNControllerReconciler) findObjectsForSrc(ctx context.Context, src client.Object) []reconcile.Request {
+	requests := []reconcile.Request{}
+
+	l := log.FromContext(ctx).WithName("Controllers").WithName("OVNController")
```
averdagu, line 255: You can use function [GetLogger](https://github.com/openstack-k8s-operators/ovn-operator/blob/main/controllers/ovncontroller_controller.go#L74), right?
Also, nit but in this file, logger is stored in a variable named "[Log](https://github.com/openstack-k8s-operators/ovn-operator/blob/main/controllers/ovncontroller_controller.go#L109)", maybe change it to conserve consistency?

templates/ovndbcluster/bin/setup.sh
```diff
@@ -26,20 +26,30 @@ fi
 if [[ "$(hostname)" != "{{ .SERVICE_NAME }}-0" ]]; then
     rm -f /etc/ovn/ovn${DB_TYPE}_db.db
     #ovsdb-tool join-cluster /etc/ovn/ovn${DB_TYPE}_db.db ${DB_NAME} tcp:$(hostname).{{ .SERVICE_NAME }}.${NAMESPACE}.svc.cluster.local:${RAFT_PORT} tcp:{{ .SERVICE_NAME }}-0.{{ .SERVICE_NAME }}.${NAMESPACE}.svc.cluster.local:${RAFT_PORT}
-    OPTS="--db-${DB_TYPE}-cluster-remote-proto=tcp --db-${DB_TYPE}-cluster-remote-addr={{ .SERVICE_NAME }}-0.{{ .SERVICE_NAME }}.${NAMESPACE}.svc.cluster.local --db-${DB_TYPE}-cluster-remote-port=${RAFT_PORT}"
+    OPTS="--db-${DB_TYPE}-cluster-remote-addr={{ .SERVICE_NAME }}-0.{{ .SERVICE_NAME }}.openstack.svc.cluster.local --db-${DB_TYPE}-cluster-remote-port=${RAFT_PORT}"
```
stuggi, line 29: `openstack` -> should be `${NAMESPACE}` ?

olliewalsh: well spotted

olliewalsh:
recheck

stuggi:
lgtm @averdagu @booxter @slawqo @karelyatin if you are also good with it, lets get it in

olliewalsh:
recheck

averdagu:
Looks good to me, let's see if yatin can also take a look as I'm not familiar with tls

pkg/ovncontroller/daemonset.go
```diff
@@ -32,6 +38,35 @@ func DaemonSet(
...
+	}
+
+	ovnControllerVolumeMounts := append(GetOvnControllerVolumeMounts(), commonVolumeMounts...)
+
+	// add OVN dbs cert and CA
+	var ovnControllerTLSArgs []string
+	if instance.Spec.TLS.Enabled() {
```
karelyatin, line 54: Wondering for setting these client config shouldn't we check if tls enabled for ovndbs? or that's too much

pkg/ovncontroller/daemonset.go
```diff
@@ -32,6 +38,35 @@ func DaemonSet(
...
+		svc := tls.Service{
+			SecretName: *instance.Spec.TLS.GenericService.SecretName,
+			CertMount:  ptr.To(ovn_common.OVNDbCertPath),
+			KeyMount:   ptr.To(ovn_common.OVNDbKeyPath),
+			CaMount:    ptr.To(ovn_common.OVNDbCaCertPath),
+		}
+		volumes = append(volumes, svc.CreateVolume(ovnv1.ServiceNameOvnController))
+		ovnControllerVolumeMounts = append(ovnControllerVolumeMounts, svc.CreateVolumeMounts(ovnv1.ServiceNameOvnController)...)
+		ovnControllerTLSArgs = []string{
+			fmt.Sprintf("--certificate==%s", ovn_common.OVNDbCertPath),
```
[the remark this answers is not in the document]
olliewalsh: https://github.com/openstack-k8s-operators/ovn-operator/pull/234/files#diff-c9fab2e2ba5d60f621f961e17a9e2bc28c84c2aa793b9849af0320188a4121edR3

pkg/ovncontroller/daemonset.go
olliewalsh: you mean check that it's `ssl:...`? Not sure if it would do anything. Ovndbs should be reconciling at the same time so it would probably just add another reconcile loop here. Can follow-up with a patch to do this later maybe if it solves a real problem

olliewalsh:
> oveall lgtm, just some missings inline.
> 
> Also wondering is something still left to get it enabled by default in a zuul job?

For OVN I need to land all of the other PRs - edpm, neutron, openstack-operator...
Think there are a few other services the need PR to land too. Then we can flip the big switch to use TLSe by default

pkg/ovncontroller/daemonset.go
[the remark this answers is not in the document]
karelyatin: seems i was unclear, why "==" , it should be just single "=" :)

pkg/ovncontroller/daemonset.go
karelyatin: Actually I was thinking of a case where tls flag get's enabled for northd/ovncontroller but not for ovndbs. May be not a real use case going forward?

olliewalsh: ah, ok, openstack-operator is managing all of this so not a real use case

olliewalsh: i.e https://github.com/openstack-k8s-operators/openstack-operator/compare/main...olliewalsh:openstack-operator:ovn_tls#diff-6dd9f181e1fc64412b7bb7a039ccfa76b7250b3d55409457fa8eb290372a4be2

karelyatin (commented):
/lgtm

PR #238, opened 2024-02-29
Add octavia mapping if present (example)

beagles:
A workaround for adding the octavia attachment as a bridge mapping if present.

(Sample patch that illustrates an intrustive yet basic approach)


beagles:
Yatin pointed out that this isn't necessary as the nicMappings will do what I want if the network attachment is already defined.

PR #241, opened 2024-03-04
Make the scripts ipv6 compatible

rabi:
In lot of these scipts we use 0.0.0.0 which assumes that we're using ipv4. This makes the scripts ipv6 compatible.

### DIFF
--- a/controllers/ovndbcluster_controller.go
+++ b/controllers/ovndbcluster_controller.go
@@ -484,7 +484,7 @@ func (r *OVNDBClusterReconciler) reconcileNormal(ctx context.Context, instance *
 	return ctrl.Result{}, nil
 }
 
-func getPodIPv4InNetwork(ovnPod corev1.Pod, namespace string, networkAttachment string) (string, error) {
+func getPodIPInNetwork(ovnPod corev1.Pod, namespace string, networkAttachment string) (string, error) {
 	netStat, err := nad.GetNetworkStatusFromAnnotation(ovnPod.Annotations)
 	if err != nil {
 		err = fmt.Errorf("Error while getting the Network Status for pod %s: %v", ovnPod.Name, err)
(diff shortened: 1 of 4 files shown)


rabi:
recheck

controllers/ovndbcluster_controller.go
```diff
@@ -630,7 +628,7 @@ func (r *OVNDBClusterReconciler) reconcileServices(
 			}
 
 			// Currently only IPv4 is supported
```
karelyatin, line 630: Will let @averdagu comment here, as there was also comment related to IPv4 Only support here.

rabi: FYI, I've tested it and works fine with the PR using ipv6 and neutron api pod can reach the ovsdbs unlike before.

hjensas (commented):
LGTM

booxter:
/lgtm

thanks

controllers/ovndbcluster_controller.go
karelyatin: Thanks @rabi ok if it works(i assume ipv6 only metallb network tests were done else it would still have used ipv4 if that's the first ip in that list), but this particular part is not used with neutron API pods(as using coredns). It's used to configure dnsdata for the pods and those are then resolved/accessed from EDPM nodes.

rabi: > i assume ipv6 only metallb network tests were done else it would still have used ipv4 if that's the first ip in that list.

I don't know what you're saying here. With single stack ipv6, all pod networks would have ipv6 ip. We don't have support dual stack yet. 

`[stack@osp-df-4 install_yamls]$ oc get po ovsdbserver-nb-0 -o jsonpath='{.metadata.annotations.k8s\.v1\.cni\.cncf\.io\/network\-status}'
[{
    "name": "ovn-kubernetes",
    "interface": "eth0",
    "ips": [
        "fd00:abcd:0:1::1784"
    ],
    "mac": "0a:58:53:53:c3:ac",
    "default": true,
    "dns": {}
},{
    "name": "openstack/internalapi",
    "interface": "internalapi",
    "ips": [
        "fd00:bbbb::31"
    ],
    "mac": "5a:01:c4:69:9f:cb",
    "dns": {}
}]
`

karelyatin: Thanks for clearing it, the output helps. yes i meant dual stack ones.

karelyatin (approved):
/lgtm

PR #245, opened 2024-03-11
Clean redundant DNS records

averdagu:
For each replica of each DB service (NB/SB) ovn-operator will create a DNSData CR. This commit modifies this and each DB service will have each own DNSData CR (ovnsb-cluster, ovnnb-cluster) and this CR will contain all the each pod IP.

Jira: [OSPRH-3628](https://issues.redhat.com//browse/OSPRH-3628)

### DIFF
--- a/controllers/ovndbcluster_controller.go
+++ b/controllers/ovndbcluster_controller.go
@@ -734,12 +734,6 @@ func (r *OVNDBClusterReconciler) reconcileServices(
 				err = fmt.Errorf("Error while deleting service with name %s: %w", fullServiceName, err)
 				return ctrl.Result{}, err
 			}
-			// Delete DNS records for deleted services/pods
-			namespace := helper.GetBeforeObject().GetNamespace()
-			err = deleteDNSData(ctx, helper, fullServiceName, namespace)
-			if err != nil {
-				return ctrl.Result{}, err
-			}
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


api/v1beta1/ovndbcluster_types.go
```diff
@@ -34,6 +34,10 @@ const (
 	// DNSSuffix : hardcoded value on how DNSCore domain is configured
 	DNSSuffix = "cluster.local"
 	// TODO: retrieve it from environment
+	// NBDBDnsName - Northbound DNS object name
+	NBDBDnsName = "ovnnb-cluster"
```
booxter, line 38: DNS: https://google.github.io/styleguide/go/decisions#initialisms should retain the original upper case for abbreviations (unless it's unexported and at the start of the variable name, so e.g. dnsList is fine for a variable).

booxter (changes requested):
since you have to rebase, please also change the variable names; but otherwise I think it's good to go

controllers/ovndbcluster_controller.go
```diff
@@ -760,23 +755,28 @@ func (r *OVNDBClusterReconciler) reconcileServices(
...
+			dnsDataName = ovnv1.NBDBDNSName
+		}
+		// Create DNSData CR
+		err = ovndbcluster.DNSData(
+			ctx,
+			helper,
+			serviceName,
+			dnsIPsList,
+			instance,
+			dnsDataName,
```
karelyatin, line 775: Instead of dnsDataName i think we can just use serviceName and avoid all the logic above.

booxter:
recheck

averdagu:
recheck

karelyatin:
recheck cert-manager issue fixed

karelyatin (approved):
/lgtm

PR #246, opened 2024-03-11
Add --n-threads option to ovn-northd

mtomaska:
Adding `--n-threads=<[0-256]>` for the ovn-norhd option. [The --n-threads](https://man7.org/linux/man-pages/man8/ovn-northd.8.html#OPTIONS) option enables parallelization for building logical flow. 
I am not limiting the new `nThreads` to 0-256 value range since the northd will enforce that policy, see the linked man page 

Closes [OSPRH-827](https://issues.redhat.com/browse/OSPRH-827)


api/v1beta1/ovnnorthd_types.go
```diff
@@ -75,6 +75,11 @@ type OVNNorthdSpecCore struct {
 	// +operator-sdk:csv:customresourcedefinitions:type=spec
 	// TLS - Parameters related to TLS
 	TLS tls.SimpleService `json:"tls,omitempty"`
+
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=1
+	// NThreads sets number of threads used for building logical flows
```
booxter, line 81: nThreads? Or maybe also consider renaming the field into numThreads so that it's more clear what this is about.

booxter (changes requested):
I think you *should* enforce it, so that it's impossible to create a OVNNorthd CR with invalid number set.

Please also add a kuttl test (or perhaps a functional test would do.)

mtomaska:
> I think you _should_ enforce it, so that it's impossible to create a OVNNorthd CR with invalid number set

Yes, when I first opened up this code  I was thinking I would do that but later decided not to. The northd already handles allowed ranges and I think it would be better to act as _passthrough_ and let the northd be the only entity enforcing it.
northd will coerce the range to 2-256 and log warning, see [n-threads in the man page](https://man7.org/linux/man-pages/man8/ovn-northd.8.html#OPTIONS) .
Also, [source](https://github.com/ovn-org/ovn/blob/894ffe8a812ba389d3c446bfd4d7d74c79d16cfc/northd/ovn-northd.c#L1164-L1176) where OVN sets it. If for some reason OVN changes this range, this code will not be affected.

BUT that was just my reasoning, if operators follow guidlines that we should enforce it, Ill be happy to change it.


> Please also add a kuttl test (or perhaps a functional test would do.)
ACK

api/v1beta1/ovnnorthd_types.go
booxter: My problem is with the comment, because this comment gets into CRD yamls (which is documentation) where it says that the name of the API field starts with upper case N (but in reality it's nThreads, as in your json: metadata).

booxter:
@mtomaska sorry, if it's just a warning in ovn, then we can leave it unenforced. Thanks for clarifying.

api/v1beta1/ovnnorthd_types.go
mtomaska: I agree, but again that is not what I see. Other variables like the `containerImage, loglevel, networkAttachments` use lower case letter for the json key, upper for the comment description and upper for the variable name in the source code. I simply just followed the established form. 
If the requirement is to have variable in the upper case then we can follow up with another PR where we make this change accross everything.

booxter: I see your point. Yes, I think others shouldn't list names of fields in wrong caps too. But I don't want to block this patch for this.

karelyatin (approved):
/lgtm

PR #247, opened 2024-03-12
Fix OVN raft cluster recovery after destructive pods' deletion

cubeek:
There is no need to wait for another pods to be spawned first. The OVN DB pods can and should start all at a time.

Resolves: [OSPRH-654](https://issues.redhat.com//browse/OSPRH-654)


cubeek:
recheck

karelyatin:
For the current failures, need to rebase to latest

booxter:
I wonder if, instead of waiting for the cluster up, we could instead configure connections / ssl / inactivity probe using Local_Config database for ovsdb-server. This should support all we need; and if so, we could avoid postStart scripts completely. I will take a look into this direction on Monday.

booxter:
UPD: Terry updated me that Local_Config is not a thing (nothing reads it) and should probably be reverted from ovs completely at this point, since config files are now supported. So we cannot use that there.

templates/ovndbcluster/bin/cleanup.sh
```diff
@@ -20,6 +20,22 @@ DB_TYPE="{{ .DB_TYPE }}"
 if [[ "${DB_TYPE}" == "sb" ]]; then
     DB_NAME="OVN_Southbound"
 fi
+
+# TODO: this logic is dedicated for non-leaders. The assumption that -0 is
+# always the leader is invalid and a bug that should be fixed.
```
slawqo, line 25: just out of curiosity: do we have jira reported for that bug already?

slawqo:
recheck

templates/ovndbcluster/bin/cleanup.sh
booxter: I am not aware of it. We should report one.

booxter:
We may want to add a destructive pod deletion kuttl test that would validate the cluster is intact after it. I am not blocking this PR for this though. But I am going to follow up on it when I have time.

templates/ovndbcluster/bin/cleanup.sh
karelyatin: +1 to have a bug if we need any optimization here.

Reasoning wise this was done to never leave the only node in the cluster and drop the database, considering scale down replicas=0 case and all pod deleted. Once we scale back to >=1 we have same cluster started instead of new bootstrap(new cid/sid)

templates/ovndbcluster/bin/setup.sh
```diff
@@ -33,8 +38,9 @@ else
     DB_ADDR="[::]"
 fi
 
+# TODO: this logic is dedicated for non-leaders. The assumption that -0 is
```
karelyatin, line 41: It was done for initial bootstrap(add *remote* args to non-0 pod), later cluster leader movement should work and not consider -0 pod as leader.

templates/ovndbcluster/bin/setup.sh
```diff
@@ -73,4 +79,44 @@ set "$@" --ovn-${DB_TYPE}-log=-vconsole:{{ .OVN_LOG_LEVEL }}
 set "$@" --ovn-${DB_TYPE}-logfile=/dev/null
 
 # don't log to file (we already log to console)
-$@ ${OPTS} run_${DB_TYPE}_ovsdb -- -vfile:off
+$@ ${OPTS} run_${DB_TYPE}_ovsdb -- -vfile:off &
+
+# Once the database is running, we will attempt to configure db options
+CTLCMD="ovn-${DB_TYPE}ctl --no-leader-only"
+
+# TODO: this logic is dedicated for the leader, not necessarily the -0 pod. The
```
karelyatin, line 87: The reasoning here was to run the db commands only on 1 of the node and considering -0 will be always there when have replicas>=1 it was choosen.

templates/ovndbcluster/bin/setup.sh
```diff
@@ -73,4 +79,44 @@ set "$@" --ovn-${DB_TYPE}-log=-vconsole:{{ .OVN_LOG_LEVEL }}
...
+# assumption that -0 is always the leader is invalid and a bug that should be fixed.
+if [[ "$(hostname)" == "{{ .SERVICE_NAME }}-0" ]]; then
+    # The command will wait until the daemon is connected and the DB is available
+    # All following ctl invocation will use the local DB replica in the daemon
+    export OVN_${DB_TYPE^^}_DAEMON=$(${CTLCMD} --pidfile --detach)
+
+{{- if .TLS }}
+    ${CTLCMD} set-ssl {{.OVNDB_KEY_PATH}} {{.OVNDB_CERT_PATH}} {{.OVNDB_CACERT_PATH}}
+    ${CTLCMD} set-connection ${DB_SCHEME}:${DB_PORT}:0.0.0.0
```
karelyatin, line 96: if there is revision can also fix 0.0.0.0 -> ${DB_ADDR}

karelyatin (commented):
/lgtm

Thanks Ihar and Kuba

templates/ovndbcluster/bin/cleanup.sh
booxter: I think you are right, indeed applying the same logic for all pods makes it problematic, because then not a single member is left to hold onto the database file and the cluster leadership.

For the case of scaling down to zero, retaining the database may still make sense for cid/sid.

I will update the comments to explain the actual reason for this. I misinterpreted the intent of the -0 name check as leadership check, but it's just a check for a "single pod in a cluster, even if it's not a leader."

templates/ovndbcluster/bin/setup.sh
booxter: Will update the comments to reflect the reasoning.

templates/ovndbcluster/bin/setup.sh
booxter: Ack, will update the comment. Curious: does it mean that we could do that in every pod with the same effect?

Also, doesn't `--no-leader-only` mean that, potentially, the remote will NOT be configured if pod0 is not leader when a config change for ssl or inactivity-probe was applied? (Will non-leader propagate the change to leader?)

templates/ovndbcluster/bin/setup.sh
booxter: Done. Btw I was wondering if we need that at all, since we already do

```
set "$@" --db-${DB_TYPE}-addr=${DB_ADDR}
set "$@" --db-${DB_TYPE}-port=${DB_PORT}
{{- if .TLS }}
set "$@" --ovn-${DB_TYPE}-db-ssl-key={{.OVNDB_KEY_PATH}}
set "$@" --ovn-${DB_TYPE}-db-ssl-cert={{.OVNDB_CERT_PATH}}
set "$@" --ovn-${DB_TYPE}-db-ssl-ca-cert={{.OVNDB_CACERT_PATH}}
```

above.

But decided not to touch it for now.

templates/ovndbcluster/bin/setup.sh
```diff
@@ -79,20 +79,21 @@ set "$@" --ovn-${DB_TYPE}-logfile=/dev/null
...
 if [[ "$(hostname)" == "{{ .SERVICE_NAME }}-0" ]]; then
     # The command will wait until the daemon is connected and the DB is available
     # All following ctl invocation will use the local DB replica in the daemon
-    export OVN_${DB_TYPE^^}_DAEMON=$(ovn-${DB_TYPE}ctl --pidfile --detach)
+    export OVN_${DB_TYPE^^}_DAEMON=$(${CTLCMD} --pidfile --detach)
 
 {{- if .TLS }}
-    ovn-${DB_TYPE}ctl --no-leader-only set-ssl {{.OVNDB_KEY_PATH}} {{.OVNDB_CERT_PATH}} {{.OVNDB_CACERT_PATH}}
-    ovn-${DB_TYPE}ctl --no-leader-only set-connection ${DB_SCHEME}:${DB_PORT}:0.0.0.0
+    ${CTLCMD} set-ssl {{.OVNDB_KEY_PATH}} {{.OVNDB_CERT_PATH}} {{.OVNDB_CACERT_PATH}}
```
cubeek, line 98: Is this required? The ctl commands should only connect to the local running daemon so the option has no effect. It makes sense to me only for the daemon process

pkg/ovndbcluster/statefulset.go
```diff
@@ -115,6 +115,19 @@ func StatefulSet(
...
+	// leave message to the leader. In general case, this should happen near
+	// instantly. But if the leader pod is itself down / restarting, it may take
+	// it some time to recover and start processing messages from other members.
+	// The default value of 30 seconds is sometimes not enough. In local testing,
+	// 60 seconds seems enough, but we'll take a significantly more conservative
+	// approach here and set it to 5 minutes.
+	//
+	// If the leader is not back even after 5 minutes, we'll give up
+	// nevertheless, and manual cluster recovery will be needed.
+	terminationGracePeriodSeconds := int64(300)
```
cubeek, line 129: Should this be a constant in pkg/ovndbcluster/const.go ?

templates/ovndbcluster/bin/setup.sh
booxter: I am not sure if this is needed; but I don't want to touch it (it's moved from elsewhere). I asked elsewhere why we need these since we also pass --ssl related arguments through CLI. We can follow up with cleanup if needed.

pkg/ovndbcluster/statefulset.go
booxter: It may be. *shrugs*

slawqo (approved):

booxter:
recheck

the cert-manager CRDs are not yet installed on the Kubernetes API server

templates/ovndbcluster/bin/setup.sh
karelyatin: Doing it in every pod will just duplicate the requests as write requests are propagated to leader as described in the doc https://github.com/ovn-org/ovn/blob/cb5d5bb3d8d2bed131112c42cab0d2ba5af8e53f/utilities/ovn-nbctl.8.xml#L213

karelyatin:
All PRs failing with same issue "the cert-manager CRDs are not yet installed on the Kubernetes API server", so need to wait for clearing it

templates/ovndbcluster/bin/setup.sh
booxter: @karelyatin but if pod-0 is not the leader, and it runs this setup as --no-leader-only, then there is a chance, I think, that this update will not get to the leader? just thinking if perhaps only the leader (whoever it is) should execute these commands, not necessarily pod-0. Or do we think that non-leader changes will always eventually propagate to leaders?

(This would not be a big deal if commands affected bootstrap only, but I believe configuration may change and so commands have to be applied on each startup.)

karelyatin: @booxter /me not sure about the case when the request not forwarded for the leader(may be you referring some network issue?), if it can't connect the leader i expect those db commands to fail and stop. And regarding the configuration change post bootstrap is a valid case.

booxter: @karelyatin do we assume that these commands will fail if non-leader pod-0 can't communicate to the leader? I am not so sure. I think we start our members independently, and we don't require for a pod to be alive to communicate with the leader. So I can imagine, in some cases, a scenario where pod-0 (non-leader) is started; these ctl commands are executed on pod-0; since --no-leader-only is passed, the commands communicate with the local member (pod-0), not the leader; they succeed; pod-0 attempts to propagate the update to the leader (?) but fails - network issue, or maybe some raft protocol hiccup; maybe pod-0 got detached from the cluster somehow... long story short, it can't sync inactivity probe into cluster; then the rest of pods in the cluster continue running without knowing about the new inactivity probe value. (The probe value here is just an example; other ctl commands we run with --no-leader-only are affected I think the same way.)

I am thinking that perhaps it would be safer for all members to run these commands against itself? Is there some side effect of doing it that is not desirable?

karelyatin:
recheck cert-manager

karelyatin:
recheck with workaround applied

templates/ovndbcluster/bin/setup.sh
karelyatin: << these ctl commands are executed on pod-0; since --no-leader-only is passed, the commands communicate with the local member (pod-0), not the leader; they succeed; pod-0 attempts to propagate the update to the leader (?) but fails - network issue, or maybe some raft protocol hiccup; maybe pod-0 got detached from the cluster somehow
@booxter from what i understood reading the document shared above, that for write/dbmodify operations it will forward the request to leader even if we pass --no-leader-only i.e and it will not apply locally and shouldn't succeed, i need to simulate this to confirm the behaviour. we can deal with this in follow up, for now let's merge it.
For point regarding running on all nodes I am not aware about side-effects of running the commands on all nodes.

karelyatin:
recheck new fix applied

templates/ovndbcluster/bin/cleanup.sh
```diff
@@ -20,6 +20,26 @@ DB_TYPE="{{ .DB_TYPE }}"
...
+    # wait for when the leader confirms we left the cluster
+    while true; do
+        # TODO: is there a better way to detect the cluster left state?..
+        STATUS=$(ovs-appctl -t /tmp/ovn${DB_TYPE}_db.ctl cluster/status ${DB_NAME} | grep Status: | awk -e '{print $2}')
+        if [ -z "$STATUS" -o "x$STATUS" = "xleft cluster" ]; then
+            break
+        fi
+        sleep 1
+    done
+
```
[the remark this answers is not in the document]
karelyatin: fixed and rebased

karelyatin:
recheck nova error consistency

karelyatin:
recheck nova gateway timeout

PR #250, opened 2024-03-15
Fix preStop Hooks for ovn-controller pod containers

karelyatin:
preStop Hooks currently fails with:-
ovs-ctl: exactly one non-option argument required
becaues ";sleep 2" is passed to the stop command which it do not expect.

To run additional commands we could use "bash -c" but considering the extra sleep do not help to avoid the "FailedPreStopHook" event just dropping the extra sleep as part of this patch.

"FailedPreStopHook" event gets raised even with successful graceful shutdown because container get's killed due to these stop commands.

To avoid "FailedPreStopHook" one option is to use wrapper startup script and handle SIGTERM signals, something like revert of [1], but leaving that for later.

Also add kuttl test to validate graceful shutdown of ovn-controller by checking SB DB for the chassis.

[1] https://github.com/openstack-k8s-operators/ovs-operator/pull/53

Resolves: [OSPRH-4553](https://issues.redhat.com//browse/OSPRH-4553)

### DIFF
--- a/tests/kuttl/tests/ovn_config/03-assert.yaml
+++ b/tests/kuttl/tests/ovn_config/03-assert.yaml
@@ -1,0 +1,13 @@
+#
+# Check for:
+#
+# - chassis unregistered in sb db
+
+apiVersion: kuttl.dev/v1beta1
+kind: TestAssert
+timeout: 30
+commands:
+    - script: |
+        sb_pod=$(oc get pod -n $NAMESPACE -l service=ovsdbserver-sb -o name|head -1)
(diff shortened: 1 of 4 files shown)


tests/kuttl/tests/ovn_config/03-assert.yaml
```diff
@@ -0,0 +1,13 @@
...
+# Check for:
+#
+# - chassis unregistered in sb db
+
+apiVersion: kuttl.dev/v1beta1
+kind: TestAssert
+timeout: 30
+commands:
+    - script: |
+        sb_pod=$(oc get pod -n $NAMESPACE -l service=ovsdbserver-sb -o name|head -1)
```
booxter, line 11: nit: spaces around | would be helpful to reader

PR #252, opened 2024-03-19
Re-init conditions each reconcile

dprince:
Also, includes ObservedGeneration Support

Jira: [OSPRH-5698](https://issues.redhat.com//browse/OSPRH-5698)
Jira: [OSPRH-5919](https://issues.redhat.com//browse/OSPRH-5919)


controllers/ovncontroller_controller.go
```diff
@@ -576,6 +571,13 @@ func (r *OVNControllerReconciler) reconcileNormal(ctx context.Context, instance
 
 	if instance.Status.NumberReady == instance.Status.DesiredNumberScheduled {
 		instance.Status.Conditions.MarkTrue(condition.DeploymentReadyCondition, condition.DeploymentReadyMessage)
+	} else {
+		instance.Status.Conditions.Set(condition.FalseCondition(
+			condition.DeploymentReadyCondition,
+			condition.RequestedReason,
+			condition.SeverityInfo,
+			condition.DeploymentReadyRunningMessage))
+		return ctrl.Result{RequeueAfter: time.Second * 10}, nil
```
gibizer, line 580: I don't think we need to actively poll this. The controller owns the dameonset it creates  
https://github.com/openstack-k8s-operators/ovn-operator/blob/cfd21c1957b75c1e1d95c1de35203a2f3e47c0a0/controllers/ovncontroller_controller.go#L239 so it will be reconciled automatically when the status of the daemonset changes.

controllers/ovncontroller_controller.go
```diff
@@ -169,25 +169,20 @@ func (r *OVNControllerReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
+		condition.UnknownCondition(condition.DeploymentReadyCondition, condition.InitReason, condition.DeploymentReadyInitMessage),
+		condition.UnknownCondition(condition.ServiceAccountReadyCondition, condition.InitReason, condition.ServiceAccountReadyInitMessage),
+		condition.UnknownCondition(condition.RoleReadyCondition, condition.InitReason, condition.RoleReadyInitMessage),
+		condition.UnknownCondition(condition.RoleBindingReadyCondition, condition.InitReason, condition.RoleBindingReadyInitMessage),
+		condition.UnknownCondition(condition.TLSInputReadyCondition, condition.InitReason, condition.InputReadyInitMessage),
+	)
 
-		// Register overall status immediately to have an early feedback e.g. in the cli
-		return ctrl.Result{}, nil
-	}
```
gibizer, line 190: We loose this behavior now. This means there can be a bit bigger time window where the Status.Conditions on the CR is empty after create. For me this is an acceptable loss. But we need to be aware of it as it might result in visible failures of pre-existing bugs in condition handling.

gibizer:
Do we have a Jira tracking similar work in all of the operators?

controllers/ovncontroller_controller.go
dprince: agree, I did not intend to leave this bit in from an earlier test

dprince:
> Do we have a Jira tracking similar work in all of the operators?

see updated description for Jira link

api/v1beta1/ovncontroller_types.go
```diff
@@ -105,6 +105,8 @@ type OVNControllerStatus struct {
 
 	// NetworkAttachments status of the deployment pods
 	NetworkAttachments map[string][]string `json:"networkAttachments,omitempty"`
+
+	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
```
ASBishop, line 109: The "Observed" prefix seems to suggest a specific meaning, but I'm not sure I understand. Would "ReconciledGeneration" be closer to what we're trying to convey?

Also, do you really want omitempty?

controllers/ovncontroller_controller.go
```diff
@@ -571,6 +571,9 @@ func (r *OVNControllerReconciler) reconcileNormal(ctx context.Context, instance
 
 	if instance.Status.NumberReady == instance.Status.DesiredNumberScheduled {
 		instance.Status.Conditions.MarkTrue(condition.DeploymentReadyCondition, condition.DeploymentReadyMessage)
+		if instance.Status.ObservedGeneration != instance.Generation {
```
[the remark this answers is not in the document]
fmount: @dprince also (I think you're already aware) this condition should be evaluated in the `openstack-operator` before `isReady` of a service operator.

api/v1beta1/ovncontroller_types.go
dprince: This is modelled after the ObservedGeneration pattern in k8s. (so I reused the name here). We can name it whatever we want I guess but it made sense to keep it similar.

Observed does mean Reconciled.

https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/deployment/deployment_controller.go#L616

dprince: re. omitempty, I guess if it hasn't reconciled it just won't show up in the status which seemed correct to me.

ASBishop: Ack (to both responses). If ObservedGeneration is a k8s thing then it makes sense to follow.

controllers/ovncontroller_controller.go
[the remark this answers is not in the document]
dprince: I removed the extra check in the latest PR. And ack on how to use this in the openstack-operator

fmount (approved):
/lgtm

abays (approved):
/lgtm

PR #258, opened 2024-04-03
Retrieve DNS domain name from cluster

averdagu:
Jira: [OSPRH-3627](https://issues.redhat.com//browse/OSPRH-3627)

### DIFF
--- a/controllers/ovndbcluster_controller.go
+++ b/controllers/ovndbcluster_controller.go
@@ -37,6 +37,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
+	operatorv1 "github.com/openshift/api/operator/v1"
 	infranetworkv1 "github.com/openstack-k8s-operators/infra-operator/apis/network/v1beta1"
 	"github.com/openstack-k8s-operators/lib-common/modules/common"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/condition"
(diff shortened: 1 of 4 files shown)


controllers/ovndbcluster_controller.go
```diff
@@ -617,6 +620,23 @@ func (r *OVNDBClusterReconciler) reconcileNormal(ctx context.Context, instance *
...
+	if err != nil {
+		Log.Info(fmt.Sprintf("Warning: Couldn't retrieve DNS cluster info, using default DNS Suffix: %s", DNSDomain))
+		return DNSDomain
+	}
+	// Using this approach in case CP have multiple domains
+	// also it does not depend on the DNS CR name.
+	// ATM in case of multiple domains will return last one.
+	for _, dns := range DNSClusterInfoList.Items {
+		DNSDomain = dns.Status.ClusterDomain
+	}
```
averdagu, line 636: I don't know if it's possible or if it'll be a use case the fact that exists multiple DNS domains (besides `cluster.local`) and how OVN operator should handle it

averdagu: Looking at the Reconcile loop of the openshift dns-operator, it only expects a DNS CR named `Default`, Could modify this chunk of code to get the object by name instead of getting a list and iterating over it.
https://github.com/openshift/cluster-dns-operator/blob/master/pkg/operator/controller/controller.go#L180-L184

karelyatin:
thx Arnau for the patch, now seeing it's usage i see it's hard coded at multiple places, so would be good to consider raising it and have a common solution and have something in lib-common/openstack-operator ?
https://github.com/search?q=org%3Aopenstack-k8s-operators%20cluster.local&type=code

averdagu:
Oh, thanks yatin for raising this, I was only focus on ovn-operator. Indeed I think it should be done globally. Putting this PR back to draft and looking into a more global solution

averdagu:
This is being handles by PR #376

PR #266, opened 2024-04-11
Switch to tempest role for running tempest

raukadah:
https://github.com/openstack-k8s-operators/ci-framework/pull/1391 moves to test_operator role for running tempest tests. It broke the existing vars based on tempest role.
It also needs to be migrated to test_operator.
Let use tempest role here to bring back running tempest tests.

### DIFF
--- a/zuul.d/jobs.yaml
+++ b/zuul.d/jobs.yaml
@@ -4,5 +4,6 @@
     parent: neutron-operator-tempest-multinode
     dependencies: ["openstack-k8s-operators-content-provider"]
     vars:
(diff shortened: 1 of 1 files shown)


karelyatin:
this shouldn't be needed as neutron-operator PR https://github.com/openstack-k8s-operators/neutron-operator/pull/333 should take care of it, and later switch to test_operator in neutron operator will auto switch this job to test_operator with more coverage, can fix the include var later to reduce test runs.

raukadah:
Thank you @karelyatin for the review. yes correct it is no longer needed!

PR #267, opened 2024-04-12
Remove the double update when finalizer added

gibizer:
The northd and dbcluster controller did a separate client.Update after
it added the self finalizer then returned and run the defer call that
did another client.Patch. However the first client.Update only wrote the
main resource and ignored the status subresource. However it refreshed
the whole instance with the result of the Update meaning it reset the in
memory status updates. We don't need two separate update so this PR
removes the standalone Update and rely on the deferred Patch call to
do all the update.

### DIFF
--- a/controllers/ovndbcluster_controller.go
+++ b/controllers/ovndbcluster_controller.go
@@ -190,6 +190,11 @@ func (r *OVNDBClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		}
 	}()
 
+	// If we're not deleting this and the service object doesn't have our finalizer, add it.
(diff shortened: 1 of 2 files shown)


booxter:
/lgtm

indeed, the client will just override the whole thing from response, regardless of which subresources were touched: https://github.com/kubernetes/client-go/blob/master/rest/request.go#L1360

gibizer:

```
failed to wait for the created cluster claim to become ready: timed out waiting for the condition
```

booxter:
recheck

```
failed to wait for the created cluster claim to become ready: timed out waiting for the condition
```

gibizer:


```
{  failed to wait for the created cluster claim to become ready: timed out waiting for the condition}

```

PR #269, opened 2024-04-19
Remove networkAttachment for ovnNorthd

umago:
ovnNorthd does not require additional interface via networkAttachment, this patch removes it from the API and controller code.

Patches removing it from the docs and sample files:

* https://github.com/openstack-k8s-operators/openstack-operator/pull/759
* https://github.com/openstack-k8s-operators/data-plane-adoption/pull/408
* https://github.com/openstack-k8s-operators/architecture/pull/186

Closes: [OSPRH-659](https://issues.redhat.com//browse/OSPRH-659)


umago:
recheck

controllers/ovnnorthd_controller.go
```diff
@@ -439,7 +399,7 @@ func (r *OVNNorthdReconciler) reconcileNormal(ctx context.Context, instance *ovn
 
 	// Define a new Deployment object
 	depl := deployment.NewDeployment(
-		ovnnorthd.Deployment(instance, serviceLabels, serviceAnnotations, nbEndpoint, sbEndpoint, envVars),
+		ovnnorthd.Deployment(instance, serviceLabels, map[string]string{}, nbEndpoint, sbEndpoint, envVars),
```
booxter, line 402: this argument is no longer needed, so can be cleaned up I think.

tests/functional/ovnnorthd_controller_test.go
```diff
@@ -335,7 +197,7 @@ var _ = Describe("OVNNorthd controller", func() {
 				Name:      "ovn-northd",
 			}
 			th.SimulateDeploymentReadyWithPods(
-				deploymentName, map[string][]string{namespace + "/internalapi": {"10.0.0.1"}},
```
booxter, line 338: This crashes in test run; I think because the controller no longer fills in the annotations, so they stay empty after create. Then lib-common assumes that SOME annotations are defined, and tries to write to `nil` map.

I think it's probably a bug in lib-common helper (it shouldn't assume IP addresses); we can fix it in lib-common, but in the meantime, you can also carry your own simulate helper that would not access nil-annotations.

If you go the in-house workaround path, please make sure you link to the PR / issue reported for lib-common, so that we can clean up later.

booxter (changes requested):

booxter:
```
vet: tests/functional/ovnnorthd_controller_test.go:28:2: "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1" imported as networkv1 and not used
```

umago:
> ```
> vet: tests/functional/ovnnorthd_controller_test.go:28:2: "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1" imported as networkv1 and not used
> ```

Yeah just saw it, forgot to run pre-commit, sorry. Fixing it.

controllers/ovnnorthd_controller.go
booxter: This is not addressed.

booxter (changes requested):
There's still an argument to clean up in Deployment function.

booxter:
You'll also need to sync lib-common into the repo to get the fix.

controllers/ovnnorthd_controller.go
umago: Ah sorry! Missed this

PR #275, opened 2024-04-25
DBCluster: remove explicit status update

gibizer:
To make it easier to reason about when the instance is updated this
patch removes the explicit status update form the input hash generation
code path and let the controller rely on the deferred PatchInstance call
as the only place where the instance is persisted.

The removed code probably didn't cause any trouble as it only updated
the status subresource and all the non status update (the self finalizer)
is guarded with an explicit return from the reconciler. So no non status
update is lost due to this status update. Still this cleanup might help
avoiding future issues.


karelyatin (commented):
/lgtm

karelyatin (approved):

PR #283, opened 2024-05-12
[Kuttl] Add test for OVN db cluster pod deletions

brfrenkel:
test steps:

1. stand up 3 replicas ovn db (nb) cluster.
2. confirm pods are all up.
3. confirm that the pods established the mesh.
4. delete all pods.
5. check that new pods are respawned.
6. check that they re-established the mesh with the correct number of pods,
and all connections are good.

related Jira: OSPRH-6135


tests/kuttl/common/scripts/check_cluster_status.sh
```diff
@@ -0,0 +1,27 @@
...
+
+# Check each pod
+for pod in "${pods[@]}"; do
+    # Execute command and capture output
+    output=$(oc exec $pod -n $NAMESPACE -- bash -c "OVS_RUNDIR=/tmp ovs-appctl -t /tmp/$ctl_file cluster/status $db_name")
+
+    # Check if the pod is a cluster member and all pods are mentioned in the connections
+    if ! echo "$output" | grep -q "Status: cluster member" || { for server in "${pods[@]}"; do ! echo "$output" | grep -q "$server" && exit 1; done; }; then
+        all_good=false
+        break
```
booxter, line 18: can't you just exit 1 here? Then you can remove `all_good` flag.

tests/kuttl/common/scripts/check_cluster_status.sh
```diff
@@ -0,0 +1,27 @@
...
+    fi
+done
+
+# Return final status
```
booxter, line 22: Just `exit 0`? (Or I believe it's the default behavior, so probably no need to do anything at all?)

tests/kuttl/common/scripts/check_cluster_status.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+
+# List of Northbound pods
+pods=("ovsdbserver-nb-0" "ovsdbserver-nb-1" "ovsdbserver-nb-2")
```
booxter, line 4: This seems unnecessarily hardcoded. I suggest you make the script accept an argument that would specify the number of expected pods, (and the type of the database.)

tests/kuttl/common/scripts/check_cluster_status.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+
+# List of Northbound pods
```
booxter, line 3: I think we should repeat the scenario for SB database too. This script should be common to both cases. Just pass the database type (sb / nb) as an argument.

tests/kuttl/common/scripts/check_cluster_status.sh
```diff
@@ -0,0 +1,27 @@
...
+
+all_good=true  # Initialize status flag
+
+# Check each pod
+for pod in "${pods[@]}"; do
+    # Execute command and capture output
+    output=$(oc exec $pod -n $NAMESPACE -- bash -c "OVS_RUNDIR=/tmp ovs-appctl -t /tmp/$ctl_file cluster/status $db_name")
+
+    # Check if the pod is a cluster member and all pods are mentioned in the connections
+    if ! echo "$output" | grep -q "Status: cluster member" || { for server in "${pods[@]}"; do ! echo "$output" | grep -q "$server" && exit 1; done; }; then
```
booxter, line 16: Could you please include an example of the string that we receive, in a comment, so that the reader understands what is being parsed?

booxter, line 16: I'm confused by this conditional, can we please split it somehow and give sub-results some meaningful names, so that it's clear for the reader what is what?

tests/kuttl/tests/ovn_db_delete/02-scale-ovndb-nb.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+commands:
+  - script: |
+      oc patch OVNDBCluster -n $NAMESPACE ovndbcluster-nb-sample --type='json' -p='[{"op": "replace", "path": "/spec/replicas", "value":3}]'
```
booxter, line 5: we should have another step for SB (to make sure it also keeps RAFT intact).

booxter (changes requested):

tests/kuttl/common/scripts/check_cluster_status.sh
brfrenkel: I changed all this section so it will be more readable, 
so it also changed :)

tests/kuttl/common/scripts/check_cluster_status.sh
brfrenkel: same as above ^

tests/kuttl/common/scripts/check_cluster_status.sh
brfrenkel: Great point! I added this :+1:

tests/kuttl/common/scripts/check_cluster_status.sh
brfrenkel: I added an example of the output that we're parsing, 
and break the condition to some parts so it'll be readable.
thanks!

brfrenkel:
@booxter thanks for the great points! 
I made the changes and now it looks much better :)

tests/kuttl/common/scripts/check_cluster_status.sh
```diff
@@ -0,0 +1,61 @@
...
+    # Status: cluster member
+    # Connections: ->0000 ->5476 <-5476 <-36cf
+    # Disconnections: 0
+    # Servers:
+    #     36cf (36cf at ssl:ovsdbserver-nb-0.ovsdbserver-nb.openstack.svc.cluster.local:6643) last msg 590 ms ago
+    #     85de (85de at ssl:ovsdbserver-nb-2.ovsdbserver-nb.openstack.svc.cluster.local:6643) (self)
+    #     5476 (5476 at ssl:ovsdbserver-nb-1.ovsdbserver-nb.openstack.svc.cluster.local:6643) last msg 12063993 ms ago
+
+    # Check if the pod is a cluster member and all pods are mentioned in the connections
+    is_cluster_member=$(echo "$output" | grep -q "Status: cluster member"; echo $?)
```
booxter, line 44: you can probably exit here if it's not 0 without running the other checks?

tests/kuttl/common/scripts/check_cluster_status.sh
```diff
@@ -0,0 +1,61 @@
...
+
+    all_pods_mentioned=true
+
+    for server in "${pods[@]}"; do
+    echo "Checking if $server is mentioned in the output"
+    if ! echo "$output" | grep -q "$server"; then
+        all_pods_mentioned=false
```
booxter, line 51: same here, I think you can just exit 1 here and then you don't need to maintain the `all_pods_mentioned` flag.

tests/kuttl/common/scripts/check_cluster_status.sh
```diff
@@ -0,0 +1,61 @@
...
+        break
+    fi
+    done
+done
+
+if [ $is_cluster_member -ne 0 ] || [ "$all_pods_mentioned" = false ]; then
```
booxter, line 57: if you `exit 1` above as suggested, this whole `if-else-fi` won't be needed I think.

booxter (changes requested):
This is almost ready. Just a minor cleanup of the script logic, but otherwise this is great!

tests/kuttl/common/scripts/check_cluster_status.sh
brfrenkel: Right :) 
I fixed those.

tests/kuttl/common/scripts/check_cluster_status.sh
```diff
@@ -0,0 +1,56 @@
...
+    for server in "${pods[@]}"; do
+    echo "Checking if $server is mentioned in the output"
+    if ! echo "$output" | grep -q "$server"; then
+        exit 1
+    fi
+    done
+done
+
+# all checks passed
+exit 0
```
booxter, line 56: nit: this is not really needed; 0 is default.

booxter:
/hold

waiting for post-beta floodgates to open

slawqo (commented):
One small idea, but don't need to be added in this patch, and maybe don't makes sense at all, that's also fine :) Maybe before 05-delete-pods step You could also add step to scale down to 1 replica (or 2), assert that this went fine and then do the delete-pods step. Please let me know what do You think about it.

Except that this patch looks good for me.

tests/kuttl/tests/ovn_db_delete/05-delete-pods.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
```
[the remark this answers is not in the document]
booxter: what do you mean? what is non-graceful in your view? killing PIDs inside containers? something else?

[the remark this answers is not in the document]
karelyatin: force delete with grace period=0 so cluster/leave do not run

booxter:
@slawqo yes I think it's a good idea to validate that delete pod in a single member cluster is also working. May be included here or as a follow-up.

tests/kuttl/tests/ovn_db_delete/05-delete-pods.yaml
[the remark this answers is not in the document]
booxter: I think we forgot to communicate this to @brfrenkel then. This can be another PR on top of this one, or included here.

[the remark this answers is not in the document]
karelyatin: No the above comment does't look related to the non-graceful delete. There may be some issue which resulted into pods crashing.

Also i see you have wrong labels used, app=ovsdbserver-nb -> service=ovsdbserver-nb

tests/kuttl/tests/ovn_db_delete/02-scale-ovndb-nb.yaml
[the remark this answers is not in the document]
brfrenkel: fixed

tests/kuttl/tests/ovn_db_delete/05-delete-pods.yaml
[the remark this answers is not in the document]
brfrenkel: Ok, as I checked with @karelyatin it is probably a local problem in my env.
I pushed a fix with non-graceful delete, and we will check the results here.

tests/kuttl/tests/ovn_db_delete/04-assert.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestAssert
```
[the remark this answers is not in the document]
brfrenkel: done

tests/kuttl/tests/ovn_db_delete/04-assert.yaml
```diff
@@ -0,0 +1,9 @@
...
+commands:
+  - script: |
+      $OVN_KUTTL_DIR/../common/scripts/check_cluster_status.sh nb 3
+      test $? -eq 0
+  - script: |
+      $OVN_KUTTL_DIR/../common/scripts/check_cluster_status.sh sb 3
```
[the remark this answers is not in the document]
brfrenkel: I used it for the scale down too, just pass 1 like:
`check_cluster_status.sh sb 1`
I think that's enough for now :)

brfrenkel:
recheck

PR #297, opened 2024-05-28
[OVNDBCluster] Check bound error on reconcileService

averdagu:
Related-Issue: [OSPRH-6798](https://issues.redhat.com//browse/OSPRH-6798)

### DIFF
--- a/controllers/ovndbcluster_controller.go
+++ b/controllers/ovndbcluster_controller.go
@@ -719,7 +719,9 @@ func (r *OVNDBClusterReconciler) reconcileServices(
 	// cluster member so it can be resolved from outside cluster (edpm nodes)
 	if instance.Spec.NetworkAttachment != "" {
 		var dnsIPsList []string
-		for _, ovnPod := range podList.Items[:*(instance.Spec.Replicas)] {
+		// TODO(averdagu): use built in Min once go1.21 is used
+		minLen := ovn_common.Min(len(podList.Items), int(*(instance.Spec.Replicas)))
+		for _, ovnPod := range podList.Items[:minLen] {
 			svc, err = service.GetServiceWithName(
 				ctx,
 				helper,
@@ -737,7 +739,9 @@ func (r *OVNDBClusterReconciler) reconcileServices(
 			}
 
 		}
-		// Create DNSData CR
+		// DNSData info is called every reconcile loop to ensure that even if a pod gets
+		// restarted and it's IP has changed, the DNSData CR will have the correct info.
+		// If nothing changed this won't modify the current dnsmasq pod.
 		err = ovndbcluster.DNSData(
 			ctx,
 			helper,
@@ -748,6 +752,14 @@ func (r *OVNDBClusterReconciler) reconcileServices(
 		)
 		if err != nil {
 			return ctrl.Result{}, err
+		}
+		// It can be possible that not all pods are ready, so DNSData won't
+		// have complete information, return error to retrigger reconcile loop
+		// Returning here instead of at the beggining of the for is done to
+		// expose the already created pods to other services/dataplane nodes
+		if len(podList.Items) < int(*(instance.Spec.Replicas)) {
+			err = fmt.Errorf("not all pods are yet created, number of expected pods: %v, current pods: %v", *(instance.Spec.Replicas), len(podList.Items))
+			return ctrl.Result{RequeueAfter: 1 * time.Second}, err
 		}
 	}
 	// dbAddress will contain ovsdbserver-(nb|sb).openstack.svc or empty
--- a/pkg/common/utils.go
+++ b/pkg/common/utils.go
@@ -1,0 +1,22 @@
+/*
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package common
+
+// Helper function while can't use buildin min
+// TOOD(averdagu) remove when using go1.21.0
+func Min(x, y int) int {
+	if x < y {
+		return x
+	}
+	return y
+}
--- a/tests/functional/ovndbcluster_controller_test.go
+++ b/tests/functional/ovndbcluster_controller_test.go
@@ -212,6 +212,47 @@ var _ = Describe("OVNDBCluster controller", func() {
 		})
 	})
 
+	When("OVNDBClusters are created with networkAttachments", func() {
+		It("does not break if pods are not created yet", func() {
+			// Create OVNDBCluster with 1 replica
+			spec := GetDefaultOVNDBClusterSpec()
+			spec.NetworkAttachment = "internalapi"
+			internalAPINADName := types.NamespacedName{Namespace: namespace, Name: "internalapi"}
+			nad := th.CreateNetworkAttachmentDefinition(internalAPINADName)
+			DeferCleanup(th.DeleteInstance, nad)
+			dbs := CreateOVNDBClusters(namespace, map[string][]string{namespace + "/internalapi": {"10.0.0.1"}}, 1)
+
+			// Increase replica to 3
+			Eventually(func(g Gomega) {
+				c := GetOVNDBCluster(dbs[0])
+				*c.Spec.Replicas = 3
+				g.Expect(k8sClient.Update(ctx, c)).Should(Succeed())
+			}).Should(Succeed())
+
+			//Check that error occurs
+			Eventually(func(g Gomega) {
+				conditions := GetOVNDBCluster(dbs[0]).Status.Conditions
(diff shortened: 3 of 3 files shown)


controllers/ovndbcluster_controller.go
```diff
@@ -719,7 +719,11 @@ func (r *OVNDBClusterReconciler) reconcileServices(
 	// cluster member so it can be resolved from outside cluster (edpm nodes)
 	if instance.Spec.NetworkAttachment != "" {
 		var dnsIPsList []string
-		for _, ovnPod := range podList.Items[:*(instance.Spec.Replicas)] {
+		for it, ovnPod := range podList.Items[:*(instance.Spec.Replicas)] {
```
booxter, line 722: I don't think you should access podList.Items past its length, even if it doesn't crash at this point.

Can't you compare lengths before iteration?

booxter: and if you'd like to iterate through the list of "pods that are supposed to run", you could iterate over `podList.Items[:min(Replicas, len(podList.Items))]` (that's pseudo-code).

averdagu: Done

controllers/ovndbcluster_controller.go
```diff
@@ -719,6 +719,10 @@ func (r *OVNDBClusterReconciler) reconcileServices(
 	// cluster member so it can be resolved from outside cluster (edpm nodes)
 	if instance.Spec.NetworkAttachment != "" {
 		var dnsIPsList []string
+		if len(podList.Items) < int(*(instance.Spec.Replicas)) {
```
booxter, line 722: This code will bail out without creating any DNSData records.

An argument can be made that it's better to create DNS records for pods that are already provisioned, even if not all of them are provisioned, yet. (so e.g. if there are 5 pods; but replicas is 7, it would be better to create DNSData records for the 5 pods and handle the rest later, on the next iteration.)

What do you think?

averdagu: Yep, I suppose that it better to have something that is slightly working than something that is not working at all. The important thing is to return error to ensure that the reconcile loop is triggered again.
The only "downside" is that the DNSData will be created every time (Can't do some check that if DNSData exist to skip it, since the fact that on every reconcile I recreate the DNSData handles the use case where pod gets restarted with a different IP)

averdagu: Adding comment regarding implementation, Need to cast all values from int to float as [Min](https://pkg.go.dev/math#Min) only accepts floats.
Also golang has a buildin [min](https://pkg.go.dev/builtin#min) function, but it was added in go1.21, currently on [go.mod](https://github.com/openstack-k8s-operators/ovn-operator/blob/main/go.mod#L3) it's defined that we use go1.20. Code not pretty but should do the work. I'm open if there's a better way.

controllers/ovndbcluster_controller.go
```diff
@@ -719,7 +720,8 @@ func (r *OVNDBClusterReconciler) reconcileServices(
 	// cluster member so it can be resolved from outside cluster (edpm nodes)
 	if instance.Spec.NetworkAttachment != "" {
 		var dnsIPsList []string
-		for _, ovnPod := range podList.Items[:*(instance.Spec.Replicas)] {
+        // TODO(averdagu): change to built in min once go1.21 is used
```
booxter, line 723: indent; please put your min code into a helper function. (if we could do without float64, just `return len(...) > Replicas? Replicas: len(...)`, that would be slightly less horrible.)

controllers/ovndbcluster_controller.go
```diff
@@ -749,6 +751,12 @@ func (r *OVNDBClusterReconciler) reconcileServices(
 		if err != nil {
 			return ctrl.Result{}, err
 		}
+		// It can be possible that not all pods are ready, so DNSData won't
+		// have complete information, return error to retrigger reconcile loop
+		if len(podList.Items) < int(*(instance.Spec.Replicas)) {
+			err = fmt.Errorf("error not all pods are yet created, number of expected pods: %v, current pods: %v", *(instance.Spec.Replicas), len(podList.Items))
+			return ctrl.Result{}, err
```
booxter, line 758: should we wait just a bit before retrying? (I think you can return a delay with Result{}, eg. 1 second?)

booxter (changes requested):

pkg/common/utils.go
```diff
@@ -0,0 +1,21 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package common
+
+// Helper function while can't use buildin min (iadded in go1.21.0)
+func Get_min(x, y int) int {
```
booxter, line 16: Golang doesn't use snake-variable-names... So `GetMin` or just `Min`.

pkg/common/utils.go
```diff
@@ -0,0 +1,21 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package common
+
+// Helper function while can't use buildin min (iadded in go1.21.0)
```
booxter, line 15: builtin

`iadded`?

add TODO comment to switch to official `Min` after 1.21.

controllers/ovndbcluster_controller.go
```diff
@@ -749,6 +751,12 @@ func (r *OVNDBClusterReconciler) reconcileServices(
 		if err != nil {
 			return ctrl.Result{}, err
 		}
+		// It can be possible that not all pods are ready, so DNSData won't
+		// have complete information, return error to retrigger reconcile loop
+		if len(podList.Items) < int(*(instance.Spec.Replicas)) {
+			err = fmt.Errorf("error not all pods are yet created, number of expected pods: %v, current pods: %v", *(instance.Spec.Replicas), len(podList.Items))
```
booxter, line 757: "error" in the message is not needed. (it will be logged as error level I believe.)

`fmt.Errorf('some pods are not created yet (%d expected, %d actual)')`

controllers/ovndbcluster_controller.go
```diff
@@ -719,7 +719,9 @@ func (r *OVNDBClusterReconciler) reconcileServices(
 	// cluster member so it can be resolved from outside cluster (edpm nodes)
 	if instance.Spec.NetworkAttachment != "" {
 		var dnsIPsList []string
-		for _, ovnPod := range podList.Items[:*(instance.Spec.Replicas)] {
+		// TODO(averdagu): change to built in min once go1.21 is used
+		min_len := ovn_common.Get_min(len(podList.Items), int(*(instance.Spec.Replicas)))
```
booxter, line 723: don't use snake-names. should be `minLen`. But actually, probably a better name can be used, e.g. `numActiveMemberPods` or smth else more explicit.

booxter (changes requested):
mostly stylistic changes requested; comments missing etc. Nothing major.

pkg/common/utils.go
```diff
@@ -0,0 +1,22 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package common
+
+// Helper function while can't use buildin min
```
booxter, line 15: builtin

controllers/ovndbcluster_controller.go
```diff
@@ -749,6 +751,12 @@ func (r *OVNDBClusterReconciler) reconcileServices(
 		if err != nil {
 			return ctrl.Result{}, err
 		}
+		// It can be possible that not all pods are ready, so DNSData won't
```
karelyatin, line 754: Trying to understand why can't we just move this length check below at L723 and avoid/remove the minLenth calculator?

karelyatin: Yes sorry, i meant "above". Not sure how much that helps to create DNSData for partial ip list. But just wanted to understand the reasoning for that, as this would just lead to unnecessary respawns for DNSMasq pod and if could avoid the hacks around Min checks. But if we still want to go with this i am ok

booxter: It will respawn once new pods show up, yes. That's sadly how dnsmasq is implemented right now.

averdagu: Yeah, I was not sure about the behavior regarding the reconcile when the DNSData doesn't change.
My concern was in the scenario where replicas > len(podList), that the reconcile will be triggered until condition is fulfilled, and in every reconcile loop (with the current code) the [DNSData creation will be called](https://github.com/openstack-k8s-operators/ovn-operator/blob/main/controllers/ovndbcluster_controller.go#L741), and I was afraid that it would trigger the recreation of the dnsmasq pod everytime. But that only will happen if the DNSData has different information.

I don't think there's much of an overhead regarding not waiting until all pods are ready. Having said that, I'm okay with any of the two solutions. 
WDYT?

averdagu: Comments added

karelyatin:
lgtm, can you rebase it on main branch

karelyatin (approved):
/lgtm

PR #299, opened 2024-06-07
Remove non graceful OVN DB pod delete Kuttl test

lewisdenny:
We are seeing a failure with the `06-delete-pods-non-graceful.yaml` job
in OCP 4.15 Prow CI. After discussing with the ovn-operator maintainers
we decided this test is invalid. With both `--force` and
`--grace-period=0` the database lock file is not being released and this
is something that requires hands on recovery which isn't being covered in
this test case.

JIRA: https://issues.redhat.com/browse/OSPRH-7515

### DIFF
--- a/tests/kuttl/tests/ovn_db_delete/06-delete-pods-non-graceful.yaml
+++ b/tests/kuttl/tests/ovn_db_delete/06-delete-pods-non-graceful.yaml
@@ -1,7 +1,0 @@
-apiVersion: kuttl.dev/v1beta1
-kind: TestStep
-commands:
-  - script: |
-      oc delete pods -n $NAMESPACE -l service=ovsdbserver-nb --force --grace-period=0
-  - script: |
-      oc delete pods -n $NAMESPACE -l service=ovsdbserver-sb --force --grace-period=0
--- a/tests/kuttl/tests/ovn_db_delete/06-assert.yaml
+++ b/tests/kuttl/tests/ovn_db_delete/06-assert.yaml
@@ -1,1 +1,0 @@
-02-assert.yaml


lewisdenny:
```
    case.go:364: failed in step 6-delete-pods-non-graceful
    case.go:366: command "" exceeded 180 sec timeout, context deadline exceeded
```

Seems the timeout needs to be increased to add a timeout -_-, can't find any docs but [1] 

[1] https://github.com/kudobuilder/kuttl/blob/f36c6725b377b518302f67dae5b4cc321dd33f7a/crds/teststep_crd.yaml#L103

booxter:
Addition of test case for --force --grace-period 0 was a mistake. This won't kill ovsdb-server, which will leave db file locked. This scenario is invalid to test - force delete should be used only when node is completely dead, I don't think we can emulate it in kuttl.

Please just remove the whole test case instead.

tests/kuttl/tests/ovn_db_delete/06-delete-pods-non-graceful.yaml
```diff
@@ -5,3 +5,7 @@ commands:
       oc delete pods -n $NAMESPACE -l service=ovsdbserver-nb --force --grace-period=0
   - script: |
       oc delete pods -n $NAMESPACE -l service=ovsdbserver-sb --force --grace-period=0
+  - script: |
```
booxter, line 8: just remove the test case for force delete, it's invalid.

booxter (changes requested):

tests/kuttl/tests/ovn_db_delete/06-delete-pods-non-graceful.yaml
lewisdenny: ack, done, thanks for the assist @booxter <3

lewisdenny:
/retest

Seems pvc failed to terminate
```
NAME                                              STATUS        VOLUME                                  CAPACITY   ACCESS MODES   STORAGECLASS    AGE
ovndbcluster-nb-sample-etc-ovn-ovsdbserver-nb-1   Terminating   local-storage01-oko-01-wnfwc-master-1   10Gi       RWO,ROX,RWX    local-storage   4m53s
ovndbcluster-nb-sample-etc-ovn-ovsdbserver-nb-2   Terminating   local-storage04-oko-01-wnfwc-master-2   10Gi       RWO,ROX,RWX    local-storage   4m53s
``

karelyatin (approved):
/lgtm

PR #302, opened 2024-06-10
Sort nicMappings to have consistent data for ObjectHash

karelyatin:
Currently the nicMappings data passed to the job spec is not ordered and leads to different Object Hash
calculated and causes ovn-config job to be retriggered and with higher number of nicMappings the job could get triggered indefinitely.
Passing the ordered list fixes the issue.

Closes-Issue: [OSPRH-7480](https://issues.redhat.com//browse/OSPRH-7480)

### DIFF
--- a/pkg/ovncontroller/utils.go
+++ b/pkg/ovncontroller/utils.go
@@ -15,6 +15,7 @@ package ovncontroller
 import (
 	"context"
 	"fmt"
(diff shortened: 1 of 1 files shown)


booxter:
/lgtm

...but functional tests are failing, probably related to the patch

PR #305, opened 2024-06-12
Fix NAD handling on create/update

karelyatin:
Currently only NAD creation was handled for the provided
nicMappings, this patch allows update to the nads.

Added missing ownership to the managed net-attach-defs
so they get's controlled with OVNController reconciler if
deleted/updated explicitly.

Also added functional tests and fixed some log messages.

Closes-Issue: [OSPRH-6894](https://issues.redhat.com//browse/OSPRH-6894)

### DIFF
--- a/pkg/ovncontroller/network.go
+++ b/pkg/ovncontroller/network.go
@@ -25,8 +25,8 @@ import (
 	ovnv1 "github.com/openstack-k8s-operators/ovn-operator/api/v1beta1"
 )
 
-// CreateAdditionalNetworks - creates network attachement definitions based on the provided mappings
-func CreateAdditionalNetworks(
+// CreateOrUpdateAdditionalNetworks - create or update network attachment definitions based on the provided mappings
+func CreateOrUpdateAdditionalNetworks(
 	ctx context.Context,
 	h *helper.Helper,
 	instance *ovnv1.OVNController,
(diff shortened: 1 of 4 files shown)


pkg/ovncontroller/network.go
```diff
@@ -25,8 +25,8 @@ import (
 	ovnv1 "github.com/openstack-k8s-operators/ovn-operator/api/v1beta1"
 )
 
-// CreateAdditionalNetworks - creates network attachement definitions based on the provided mappings
-func CreateAdditionalNetworks(
+// CreateOrUpdateAdditionalNetworks - creates network attachement definitions based on the provided mappings
```
booxter, line 28: while you are here anyway, perhaps: `attachment`; also - should this text reflect the new goal of the function - to create **and update**?

pkg/ovncontroller/network.go
```diff
@@ -52,23 +57,36 @@ func CreateAdditionalNetworks(
...
+			for _, owner := range nad.GetOwnerReferences() {
+				if owner.Name == instance.Name {
+					owned = true
+					break
+				}
+			}
+			if owned {
+				nad.Spec = nadSpec
+				if err := h.GetClient().Update(ctx, nad); err != nil {
+					return nil, fmt.Errorf("can not update NetworkAttachmentDefinition %s/%s: %w",
```
booxter, line 86: nit: can not -> cannot

```
Don’t use can not when you mean cannot. The only time you’re likely to see can not written as separate words is when the word “can” happens to precede some other phrase that happens to start with “not”:
```

(though I see it's the same above for create case... so I dunno...)

booxter (changes requested):
Looks correct, just nits + please add func tests for the scenarios.

pkg/ovncontroller/network.go
karelyatin: Done

pkg/ovncontroller/network.go
karelyatin: Update at all places

karelyatin:
recheck quay registry issue

karelyatin:
recheck https://github.com/openstack-k8s-operators/ovn-operator/pull/306

karelyatin:
recheck https://github.com/openstack-k8s-operators/ovn-operator/pull/306

PR #313, opened 2024-06-27
Add required csv annotations. Remove deprecated

mtomaska:
Removing: 
`operators.openshift.io/infrastructure-features: ["disconnected"]` is deprecated starting with OpenShift 4.14
The `features.operators.openshift.io/disconnected: "true"` annotation is already present as its replacament

Adding newly required csv annotations. Set to "false" to indicate this operator does not support them

Implements: https://issues.redhat.com/browse/OSPRH-7931


karelyatin (approved):
/lgtm

PR #315, opened 2024-07-01
Gracefully handle empty db file on startup

booxter:
In some situations, e.g. when ovsdb-tool is killed during db file initialization, the file may end up empty. On next restart, the service fails to start because the file format is invalid.

Ideally, ovs start scripts would handle this situation gracefully, as tracked in https://issues.redhat.com/browse/FDP-689

But in the meantime, we can also handle the situation on operator's side, by removing the empty file if present.

FYI a valid file always contains at least some data, because ovsdb-tool adds a log entry with raft config as the first entry. See:

https://github.com/openvswitch/ovs/blob/56e315937eeb640d5d8f305988d133390445eaee/ovsdb/raft.c#L525

Resolves: https://issues.redhat.com/browse/OSPRH-8117

### DIFF
--- a/templates/ovndbcluster/bin/cleanup.sh
+++ b/templates/ovndbcluster/bin/cleanup.sh
@@ -14,9 +14,9 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 set -ex
+source $(dirname $0)/functions
 
 DB_NAME="OVN_Northbound"
-DB_TYPE="{{ .DB_TYPE }}"
 if [[ "${DB_TYPE}" == "sb" ]]; then
     DB_NAME="OVN_Southbound"
 fi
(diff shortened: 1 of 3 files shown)


booxter:
@karelyatin this is the fix body. Do you think we need a kuttl scenario for this? We can probably start the cluster, then `echo -n > $DB_FILE`, then kill the pod and see if it restarts successfully.

booxter:
```
 error: build error: Failed to push image: trying to reuse blob sha256:e394ea8406c7ed85ddc862f882ec77dcfd3863de3cb90fd8006238d521d22d44 at destination: pinging container registry quay.rdoproject.org: Get "https://quay.rdoproject.org/v2/": dial tcp 38.129.56.158:443: i/o timeout 
```

karelyatin:
> @karelyatin this is the fix body. Do you think we need a kuttl scenario for this? We can probably start the cluster, then `echo -n > $DB_FILE`, then kill the pod and see if it restarts successfully.

Thanks Ihar, doesn't looks we need any kuttl test just for this.

PR #320, opened 2024-07-05
Stop the OVN database during the cleanup script

ralonsoh:
Closes-Issue: [OSPRH-8118](https://issues.redhat.com//browse/OSPRH-8118)

### DIFF
--- a/templates/ovndbcluster/bin/cleanup.sh
+++ b/templates/ovndbcluster/bin/cleanup.sh
@@ -44,3 +44,9 @@ if [[ "$(hostname)" != "{{ .SERVICE_NAME }}-0" ]]; then
     # now that we left, the database file is no longer valid
     cleanup_db_file
 fi
+
+# Stop the DB server gracefully; this will also end the pod running script.
+# By executing the stop command in the background, we guarantee to exit
+# this script and not fail with "FailedPreStopHook", while the database is
+# stopped.
+/usr/share/ovn/scripts/ovn-ctl stop_${DB_TYPE}_ovsdb &


olliewalsh:
> Closes-Issue: [OSPRH-8118](https://issues.redhat.com//browse/OSPRH-8118)

How about just removing` --single-child` from the dumb-init args so it sends SIGTERM to the entire process group?

templates/ovndbcluster/bin/cleanup.sh
```diff
@@ -44,3 +44,10 @@ if [[ "$(hostname)" != "{{ .SERVICE_NAME }}-0" ]]; then
     # now that we left, the database file is no longer valid
     cleanup_db_file
 fi
+
+# Stop the DB server gracefully; this will also end the pod running script.
+if [ "$DB_TYPE" == "nb" ]; then
+    /usr/share/ovn/scripts/ovn-ctl stop_nb_ovsdb
+elif [ "$DB_TYPE" == "sb" ]; then
+    /usr/share/ovn/scripts/ovn-ctl stop_sb_ovsdb
```
booxter, line 52: you can just `stop_${DB_TYPE}_ovsdb` I think.

booxter:
@olliewalsh the problem with (just) removing `--single-child` is that `dumb-init` won't wait for children-of-children to exit, it will only wait for the start script to exit. (And the latter will exit immediately because there's no `trap` set for `SIGTERM` there to e.g. wait for dbserver to exit.) - Alternative solution could be to implement `SIGTERM` handler in the start script.

booxter:
@olliewalsh btw FYI the signal handling improvement is tracked here: https://issues.redhat.com/browse/OSPRH-8212

ralonsoh:
> @olliewalsh btw FYI the signal handling improvement is tracked here: https://issues.redhat.com/browse/OSPRH-8212

If we are going to implement this signal handler in the start script, is this patch relevant anymore?

booxter:
> If we are going to implement this signal handler in the start script, is this patch relevant anymore?

I think it depends on how the signal handler would be implemented. The handler could handle this job, or it could only take care about ovsdb-tool db initialization. (Which is the original scenario that spurred these Jira issues.)

But yes, I think everything could be handled inside the signal handler.

slawqo (commented):
/lgtm for me

booxter:
@slawqo believe ci is green here now

booxter:
I've removed lgtm for now so that it doesn't merge without first looking at the failure logs. Once this is resolved, I will add the label back.

ralonsoh:
> I've removed lgtm for now so that it doesn't merge without first looking at the failure logs. Once this is resolved, I will add the label back.

Checking the CI logs now and trying to reproduce the issue locally.

ralonsoh:
Pushing an empty commit, to test the upstream CI (I don't reproduce this issue in my dev environment)

ralonsoh:

From the logs:
` Removing debug pod ...
error: unable to upgrade connection: container container-00 not found in pod oko-18-5hpg6-master-2-debug-zqrt7_openstack
make: *** [Makefile:565: crc_storage] Error 1
+ n=3
+ ((  n >= retries  ))
+ echo 'Failed to run '\''make crc_storage'\'' target. Aborting'
Failed to run 'make crc_storage' target. Aborting
+ exit 1
{"component":"entrypoint","error":"wrapped process failed: exit status 1","file":"sigs.k8s.io/prow/pkg/entrypoint/run.go:84","func":"sigs.k8s.io/prow/pkg/entrypoint.Options.internalRun","level":"error","msg":"Error executing test process","severity":"error","time":"2024-09-20T09:25:15Z"}
error: failed to execute wrapped command: exit status 1 
INFO[2024-09-20T09:25:16Z] Step ovn-operator-build-deploy-kuttl-openstack-k8s-operators-kuttl failed after 1m9s. 
INFO[2024-09-20T09:25:16Z] Step phase test failed after 18m14s.    `

templates/ovndbcluster/bin/cleanup.sh
```diff
@@ -44,3 +44,11 @@ if [[ "$(hostname)" != "{{ .SERVICE_NAME }}-0" ]]; then
     # now that we left, the database file is no longer valid
     cleanup_db_file
 fi
+
+# Stop the DB server gracefully; this will also end the pod running script.
+# By executing the stop command in the backgroung, we guarantee to exit
+# this script and not fail with "FailedPreStopHook", while the database is
+# stopped and the pod exits.
+if [[ "$(hostname)" == "{{ .SERVICE_NAME }}-0" ]]; then
```
booxter, line 52: Could you please explain why only the -0 member is given a chance to gracefully exit? (And the rest will get SIGKILLs from kubelet, I think.)

templates/ovndbcluster/bin/cleanup.sh
```diff
@@ -44,3 +44,11 @@ if [[ "$(hostname)" != "{{ .SERVICE_NAME }}-0" ]]; then
     # now that we left, the database file is no longer valid
     cleanup_db_file
 fi
+
+# Stop the DB server gracefully; this will also end the pod running script.
```
booxter, line 48: (Side note, not necessarily the issue to resolve for this PR)

I wonder if now that we [have](https://github.com/openstack-k8s-operators/ovn-operator/pull/327/files) a proper trap handler set in the startup script, we should not move the `cleanup.sh` code there.

templates/ovndbcluster/bin/cleanup.sh
```diff
@@ -44,3 +44,11 @@ if [[ "$(hostname)" != "{{ .SERVICE_NAME }}-0" ]]; then
...
+# By executing the stop command in the backgroung, we guarantee to exit
+# this script and not fail with "FailedPreStopHook", while the database is
```
booxter, line 50: Do you mean that the PreStop script times out waiting for the `ovsdb-server` process to exit? Do you know why this is happening?

templates/ovndbcluster/bin/cleanup.sh
ralonsoh: That's probably a leftover of my previous tests. I initially thought that this script was stopping the DB server and deleting the DB file, but this is wrong: this script is just stopping the ovsdb daemon, nothing else. This should apply to any DB server.

templates/ovndbcluster/bin/cleanup.sh
ralonsoh: No, the problem is that when we stop the ovsdb server, the init scrip ends and the pod exits. That happens before we leave this script and this is why we have this "FailedPreStopHook" error. This is a cosmetic change, because that happens at the end of this script and we actually achieve what we want, that is to stop the ovsdb server.

templates/ovndbcluster/bin/cleanup.sh
```diff
@@ -44,3 +44,11 @@ if [[ "$(hostname)" != "{{ .SERVICE_NAME }}-0" ]]; then
     # now that we left, the database file is no longer valid
     cleanup_db_file
 fi
+
+# Stop the DB server gracefully; this will also end the pod running script.
+# By executing the stop command in the backgroung, we guarantee to exit
```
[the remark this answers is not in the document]
ralonsoh: Done

templates/ovndbcluster/bin/cleanup.sh
ralonsoh: Ok, with the [latest PS](https://github.com/openstack-k8s-operators/ovn-operator/pull/320/commits/ee2ec81a6d408e416e066b77ae0a00714dc56c5a), that includes all DB replicas, the kuttl tests are passing.

slawqo (approved):

PR #341, opened 2024-08-13
[ovsdb-server] Move db initialization to an init container

karelyatin:
There is a potential race condition where ovs-vswitchd containers can run ovs db commands and fails while ovsdb-server stops and restarts after db initialization. This makes ovs-vswitchd container to fail and restart and it eventually recovers and it's better to avoid this.

This patch moves the db initialization part to init container to avoid this race.

Resolves: [OSPRH-637](https://issues.redhat.com//browse/OSPRH-637)

### DIFF
--- a/pkg/ovncontroller/daemonset.go
+++ b/pkg/ovncontroller/daemonset.go
@@ -169,6 +169,24 @@ func CreateOVSDaemonSet(
 	envVars := map[string]env.Setter{}
 	envVars["CONFIG_HASH"] = env.SetValue(configHash)
 
+	initContainers := []corev1.Container{
+		{
+			Name:    "ovsdb-server-init",
+			Command: []string{"/usr/local/bin/container-scripts/init-ovsdb-server.sh"},
+			Image:   instance.Spec.OvsContainerImage,
+			SecurityContext: &corev1.SecurityContext{
+				Capabilities: &corev1.Capabilities{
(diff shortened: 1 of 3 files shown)


pkg/ovncontroller/daemonset.go
```diff
@@ -169,6 +169,24 @@ func CreateOVSDaemonSet(
...
 	envVars["CONFIG_HASH"] = env.SetValue(configHash)
 
+	initContainers := []corev1.Container{
+		{
+			Name:    "ovsdb-server-init",
+			Command: []string{"/usr/local/bin/container-scripts/init-ovsdb-server.sh"},
+			Image:   instance.Spec.OvsContainerImage,
+			SecurityContext: &corev1.SecurityContext{
+				Capabilities: &corev1.Capabilities{
+					Add:  []corev1.Capability{"NET_ADMIN", "SYS_ADMIN", "SYS_NICE"},
```
booxter, line 179: I don't know if these capabilities are all required for the command; but I guess the PR doesn't escalate anything from what it was doing without the PR, so it should be ok, even if the list could be cleaned up.

booxter (approved):

pkg/ovncontroller/daemonset.go
karelyatin: Yes i just replicated the ones already used

PR #352, opened 2024-09-25
Move external config map generation to OVNDBCluster

averdagu:
Currently the external config map consumed by edpm deployment is generated on ovn-controller. 
Due to the fact that the environment can be set without deploying ovn-controller (if nicMappings is empty there's no need to spawn it) the config map wouldn't be created. Since this config map is needed by the edpm deployment it's creation has been moved from ovn-controller to ovndbcluster. 

Resolves: [OSPRH-10372](https://issues.redhat.com//browse/OSPRH-10372)
Related-Issue: [OSPRH-7463](https://issues.redhat.com//browse/OSPRH-7463)

Resolves: https://github.com/openstack-k8s-operators/openstack-operator/pull/1070

### DIFF
--- a/api/v1beta1/client.go
+++ b/api/v1beta1/client.go
@@ -52,6 +52,26 @@ func getDBClusters(
 	return ovnDBList, nil
 }
 
+func GetOVNController(
+	ctx context.Context,
+	h *helper.Helper,
+	namespace string,
+) (*OVNController, error) {
+	ovnControllerList := &OVNControllerList{}
+	listOpts := []client.ListOption{
+		client.InNamespace(namespace),
+	}
+	err := h.GetClient().List(ctx, ovnControllerList, listOpts...)
+	if err != nil {
+		return nil, err
+	}
+	if len(ovnControllerList.Items) > 0 {
+		return &ovnControllerList.Items[0], nil
+	}
+
+	return nil, nil
+}
+
 // GetDBClusterByType - return OVNDBCluster for the given dbType
 func GetDBClusterByType(
 	ctx context.Context,
@@ -87,8 +107,8 @@ func getItems(list client.ObjectList) []client.Object {
 	return items
 }
 
-// OVNDBClusterNamespaceMapFunc - DBCluster Watch Function
-func OVNDBClusterNamespaceMapFunc(crs client.ObjectList, reader client.Reader) handler.MapFunc {
+// OVNCRNamespaceMapFunc // Generic function to watch any OVN CR
+func OVNCRNamespaceMapFunc(crs client.ObjectList, reader client.Reader) handler.MapFunc {
 	return func(ctx context.Context, obj client.Object) []reconcile.Request {
 		log := log.FromContext(ctx)
 		result := []reconcile.Request{}
--- a/controllers/ovncontroller_controller.go
+++ b/controllers/ovncontroller_controller.go
@@ -25,7 +25,6 @@ import (
 	"github.com/go-logr/logr"
 	netattdefv1 "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1"
 	k8s_errors "k8s.io/apimachinery/pkg/api/errors"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
@@ -227,7 +226,7 @@ func (r *OVNControllerReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Owns(&corev1.ServiceAccount{}).
 		Owns(&rbacv1.Role{}).
 		Owns(&rbacv1.RoleBinding{}).
-		Watches(&ovnv1.OVNDBCluster{}, handler.EnqueueRequestsFromMapFunc(ovnv1.OVNDBClusterNamespaceMapFunc(crs, mgr.GetClient()))).
+		Watches(&ovnv1.OVNDBCluster{}, handler.EnqueueRequestsFromMapFunc(ovnv1.OVNCRNamespaceMapFunc(crs, mgr.GetClient()))).
 		Watches(
 			&corev1.Secret{},
 			handler.EnqueueRequestsFromMapFunc(r.findObjectsForSrc),
@@ -601,33 +600,8 @@ func (r *OVNControllerReconciler) reconcileNormal(ctx context.Context, instance 
 
 	sbCluster, err := ovnv1.GetDBClusterByType(ctx, helper, instance.Namespace, map[string]string{}, ovnv1.SBDBType)
 	if err != nil {
-		Log.Info("No SB OVNDBCluster defined, deleting external ConfigMap")
-		cleanupConfigMapErr := r.deleteExternalConfigMaps(ctx, helper, instance)
-		if cleanupConfigMapErr != nil {
-			Log.Error(cleanupConfigMapErr, "Failed to delete external ConfigMap")
-			return ctrl.Result{}, cleanupConfigMapErr
-		}
+		Log.Info("No SB OVNDBCluster defined. Exiting reconcile.")
 		return ctrl.Result{}, nil
-	}
-
-	ep, err := sbCluster.GetExternalEndpoint()
-	if err != nil || ep == "" {
-		Log.Info("No external endpoint defined for SB OVNDBCluster, deleting external ConfigMap")
-		cleanupConfigMapErr := r.deleteExternalConfigMaps(ctx, helper, instance)
-		if cleanupConfigMapErr != nil {
-			Log.Error(cleanupConfigMapErr, "Failed to delete external ConfigMap")
-			return ctrl.Result{}, cleanupConfigMapErr
-		}
-	}
-
-	if sbCluster.Spec.NetworkAttachment != "" {
-		// Create ConfigMap for external dataplane consumption
-		// TODO(ihar) - is there any hashing mechanism for EDP config? do we trigger deploy somehow?
-		err = r.generateExternalConfigMaps(ctx, helper, instance, sbCluster, &configMapVars)
-		if err != nil {
-			Log.Error(err, "Failed to generate external ConfigMap")
-			return ctrl.Result{}, err
-		}
 	}
 
 	// create OVN Config Job - start
@@ -717,61 +691,6 @@ func (r *OVNControllerReconciler) generateServiceConfigMaps(
 		},
 	}
 	return configmap.EnsureConfigMaps(ctx, h, instance, cms, envVars)
-}
-
-// generateExternalConfigMaps - create configmaps for external dataplane consumption
-func (r *OVNControllerReconciler) generateExternalConfigMaps(
-	ctx context.Context,
-	h *helper.Helper,
-	instance *ovnv1.OVNController,
-	sbCluster *ovnv1.OVNDBCluster,
-	envVars *map[string]env.Setter,
-) error {
-	// Create/update configmaps from templates
-	cmLabels := labels.GetLabels(instance, labels.GetGroupLabel(ovnv1.ServiceNameOVNController), map[string]string{})
-
-	externalEndpoint, err := sbCluster.GetExternalEndpoint()
-	if err != nil {
-		return err
-	}
-
-	externalTemplateParameters := make(map[string]interface{})
-	externalTemplateParameters["OVNRemote"] = externalEndpoint
-	externalTemplateParameters["OVNEncapType"] = instance.Spec.ExternalIDS.OvnEncapType
-
-	cms := []util.Template{
-		// EDP ConfigMap
-		{
-			Name:          fmt.Sprintf("%s-config", instance.Name),
-			Namespace:     instance.Namespace,
-			Type:          util.TemplateTypeConfig,
-			InstanceType:  instance.Kind,
-			Labels:        cmLabels,
-			ConfigOptions: externalTemplateParameters,
-		},
-	}
-	return configmap.EnsureConfigMaps(ctx, h, instance, cms, envVars)
-}
-
-// TODO(ihar) this function could live in lib-common
-// deleteExternalConfigMaps - delete obsolete configmaps for external dataplane consumption
-func (r *OVNControllerReconciler) deleteExternalConfigMaps(
-	ctx context.Context,
-	h *helper.Helper,
-	instance *ovnv1.OVNController,
-) error {
-	cm := &corev1.ConfigMap{
-		ObjectMeta: metav1.ObjectMeta{
-			Name:      fmt.Sprintf("%s-config", instance.Name),
-			Namespace: instance.Namespace,
-		},
-	}
-
-	err := h.GetClient().Delete(ctx, cm)
-	if err != nil && !k8s_errors.IsNotFound(err) {
-		return fmt.Errorf("error deleting external config map %s: %w", cm.Name, err)
-	}
-	return nil
 }
 
 // createHashOfInputHashes - creates a hash of hashes which gets added to the resources which requires a restart
--- a/controllers/ovndbcluster_controller.go
+++ b/controllers/ovndbcluster_controller.go
@@ -23,6 +23,7 @@ import (
 	"time"
 
 	"github.com/go-logr/logr"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
@@ -86,6 +87,7 @@ func (r *OVNDBClusterReconciler) GetLogger(ctx context.Context) logr.Logger {
 }
 
 //+kubebuilder:rbac:groups=ovn.openstack.org,resources=ovndbclusters,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=ovn.openstack.org,resources=ovncontroller,verbs=watch;
 //+kubebuilder:rbac:groups=ovn.openstack.org,resources=ovndbclusters/status,verbs=get;update;patch
 //+kubebuilder:rbac:groups=ovn.openstack.org,resources=ovndbclusters/finalizers,verbs=update;patch
 //+kubebuilder:rbac:groups=core,resources=configmaps,verbs=get;list;watch;create;update;patch;delete;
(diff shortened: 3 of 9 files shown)


controllers/ovndbcluster_controller.go
```diff
@@ -786,6 +805,70 @@ func (r *OVNDBClusterReconciler) reconcileServices(
...
+
+	externalEndpoint, err := instance.GetExternalEndpoint()
+	if err != nil {
+		return err
+	}
+
+	ovnController, err := ovnv1.GetOVNController(ctx, h, instance.Namespace)
+	if k8s_errors.IsNotFound(err) {
+		// If ovnController is not found use geneve as default
+		encapType = "geneve"
```
averdagu, line 828: What should we do when OVNController is not spawned?
If there's no nicMappings configured, openstack-operator won't create any OVNController CR, so even if the OSCP has the encapType information, ovn-operator won't have the CR to get it, should we set geneve as default? Leave it empty?

averdagu (commented):
Started discussion on ovndbcluster_controller.go

controllers/ovndbcluster_controller.go
karelyatin: We do not need to set/pass it to edpm if there is no ovn controller on control plane side. ovn-controller on edpm nodes will keep using what configured there i.e with edpm_ovn_encap_type

controllers/ovndbcluster_controller.go
```diff
@@ -612,6 +622,15 @@ func (r *OVNDBClusterReconciler) reconcileNormal(ctx context.Context, instance *
 
 		// Set DB Address
 		instance.Status.InternalDBAddress = strings.Join(internalDbAddress, ",")
+		if instance.Spec.DBType == ovnv1.SBDBType {
```
karelyatin, line 625: can also check for networkattach != nil

controllers/ovncontroller_controller.go
```diff
@@ -602,34 +601,9 @@ func (r *OVNControllerReconciler) reconcileNormal(ctx context.Context, instance
 	sbCluster, err := ovnv1.GetDBClusterByType(ctx, helper, instance.Namespace, map[string]string{}, ovnv1.SBDBType)
 	if err != nil {
 		Log.Info("No SB OVNDBCluster defined, deleting external ConfigMap")
```
karelyatin, line 603: can drop external configmap delete reference from logs now

averdagu: Done

controllers/ovndbcluster_controller.go
averdagu: Done

controllers/ovndbcluster_controller.go
```diff
@@ -612,6 +622,15 @@ func (r *OVNDBClusterReconciler) reconcileNormal(ctx context.Context, instance *
...
+			// This config map will populate the sb db address to edpm, can't use the nb, can't use the nb
```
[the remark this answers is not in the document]
averdagu: Done

api/v1beta1/client.go
```diff
@@ -52,6 +53,21 @@ func getDBClusters(
 	return ovnDBList, nil
 }
 
+func GetOVNController(
+	ctx context.Context,
+	h *helper.Helper,
+	namespace string,
+) (*OVNController, error) {
+	ovnController := &OVNController{}
+	name := types.NamespacedName{Name: "ovncontroller", Namespace: namespace}
```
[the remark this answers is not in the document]
averdagu: Done, changed to list and get first item

controllers/ovndbcluster_controller.go
averdagu: Yep, either we setup 'geneve' whenever ovnController is not set (geneve is the [default in the edpm_ansible](https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/roles/edpm_ovn/defaults/main.yml#L41) ovn role) or we don't set up the encapType at all, which will need a modification on how the configMap is generated.
Whatever is set here will overwrite the default on edpm_ansible (even if it's empty)

controllers/ovncontroller_controller.go
```diff
@@ -601,35 +600,10 @@ func (r *OVNControllerReconciler) reconcileNormal(ctx context.Context, instance
...
 	sbCluster, err := ovnv1.GetDBClusterByType(ctx, helper, instance.Namespace, map[string]string{}, ovnv1.SBDBType)
 	if err != nil {
-		Log.Info("No SB OVNDBCluster defined, deleting external ConfigMap")
-		cleanupConfigMapErr := r.deleteExternalConfigMaps(ctx, helper, instance)
-		if cleanupConfigMapErr != nil {
-			Log.Error(cleanupConfigMapErr, "Failed to delete external ConfigMap")
-			return ctrl.Result{}, cleanupConfigMapErr
-		}
+		Log.Info("No SB OVNDBCluster defined. Exiting reconcile.")
 		return ctrl.Result{}, nil
```
booxter, line 604: is it correct to return "success" status when 1) there is OVNController but 2) there is no SB db? I thought that we are making sure that OVNController doesn't exist in openstack-operator; and if it does exist, then either openstack-operator is wrong, or SB down - also a problem. Should we then raise an error?

controllers/ovndbcluster_controller.go
```diff
@@ -374,6 +375,15 @@ func (r *OVNDBClusterReconciler) reconcileNormal(ctx context.Context, instance *
 				err.Error()))
 			return ctrl.Result{}, err
 		}
+	} else if instance.Spec.DBType == ovnv1.SBDBType {
+		// This config map was created by the SB and it only needs to be deleted once
+		// since this reconcile loop can be done by the SB and the NB, filtering so only
+		// one deletes it.
+		Log.Info("NetworkAttachment is empty, deleting ExternalConfigMaps")
```
booxter, line 382: `ExternalConfigMaps` what is that? It's not a class, not a variable. Should it use proper words then? "deleting external config map(s)"?

controllers/ovndbcluster_controller.go
```diff
@@ -612,6 +622,16 @@ func (r *OVNDBClusterReconciler) reconcileNormal(ctx context.Context, instance *
 
 		// Set DB Address
 		instance.Status.InternalDBAddress = strings.Join(internalDbAddress, ",")
+		if instance.Spec.DBType == ovnv1.SBDBType && instance.Spec.NetworkAttachment != "" {
+			// This config map will populate the sb db address to edpm, can't use the nb
+			// If there's no networkAttachments the configMap is not needed
```
booxter, line 627: why would we not unconditionally create the config map? I don't think anything breaks if we do, and one may want to inspect the configMap even if the SB pod is not attached to an additional network, yet. Not sure why we have to condition it.

controllers/ovndbcluster_controller.go
```diff
@@ -786,6 +806,69 @@ func (r *OVNDBClusterReconciler) reconcileServices(
...
+	if err != nil {
+		return err
+	} else if ovnController != nil {
+		encapType = ovnController.Spec.ExternalIDS.OvnEncapType
+	}
+
+	externalTemplateParameters := make(map[string]interface{})
+	externalTemplateParameters["OVNRemote"] = externalEndpoint
+	if encapType != "" {
+		externalTemplateParameters["OVNEncapType"] = encapType
```
booxter, line 836: (Not action required, Observation) Long term, I think this config may be better separated into a separate ConfigMap, then both can be mounted / consumed by edpm-ansible role.

controllers/ovndbcluster_controller.go
```diff
@@ -786,6 +806,69 @@ func (r *OVNDBClusterReconciler) reconcileServices(
...
+	}
+
+	cms := []util.Template{
+		// EDP ConfigMap
+		{
+			Name:          "ovncontroller-config",
```
booxter, line 842: (No action required, Observation) The name here is now a bit unfortunate. Yes, it's "consumed by ovncontroller" (for what we know); but it's produced by DbCluster CRD controller. If/when we split the encap-type config into a separate configMap and maybe start managing it in OVNController controller handler, then the naming may have to be revisited.

controllers/ovndbcluster_controller.go
```diff
@@ -786,6 +806,69 @@ func (r *OVNDBClusterReconciler) reconcileServices(
...
+	// Create/update configmaps from templates
+	cmLabels := labels.GetLabels(instance, labels.GetGroupLabel(serviceName), map[string]string{})
+	encapType := ""
+
+	externalEndpoint, err := instance.GetExternalEndpoint()
+	if err != nil {
+		return err
+	}
+
+	ovnController, err := ovnv1.GetOVNController(ctx, h, instance.Namespace)
```
booxter, line 826: nit: it may simplify the code a bit if you can drop `encapType` variable like:

```
externalTemplateParameters := make(map[string]interface{})
externalTemplateParameters["OVNRemote"] = externalEndpoint

ovnController, ... = ...GetOVNController(...)
if err != nil {
    return err
}
if ovnController != nil {
    externalTemplateParameters["OVNEncapType"] = ovnController.Spec.ExternalIDS.OvnEncapType
}
```

controllers/ovndbcluster_controller.go
```diff
@@ -786,6 +806,69 @@ func (r *OVNDBClusterReconciler) reconcileServices(
...
+}
+
+func (r *OVNDBClusterReconciler) deleteExternalConfigMaps(
+	ctx context.Context,
+	h *helper.Helper,
+	namespace string,
+) error {
+	cm := &corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "ovncontroller-config",
```
booxter, line 860: Before this PR, the name of the CM was constructed from the OVNController instance name + `-config` prefix. Now it's hardcoded. This will break in case OVNController CR is named differently from `ovncontroller`. Is it safe? What if a user names their CR differently?

tests/functional/ovndbcluster_controller_test.go
```diff
@@ -298,6 +328,119 @@ var _ = Describe("OVNDBCluster controller", func() {
...
+			Expect(th.GetConfigMap(externalCM).ObjectMeta.OwnerReferences[0].Name).To(Equal(OVNDBClusterName.Name))
+			Expect(th.GetConfigMap(externalCM).ObjectMeta.OwnerReferences[0].Kind).To(Equal("OVNDBCluster"))
+
+			Eventually(func(g Gomega) {
+				g.Expect(th.GetConfigMap(externalCM).Data["ovsdb-config"]).Should(
+					ContainSubstring("ovn-remote: %s", ExpectedExternalSBEndpoint))
+			}, timeout, interval).Should(Succeed())
+		})
+
+		It("should create an external ConfigMap with ovn-encap-type if OVNController is configured", func() {
```
booxter, line 366: please add a test case with OVNController named differently (not `ovncontroller`).

booxter: maybe check that removal of OVNController resets the CM to not include encap type?

booxter (changes requested):

controllers/ovndbcluster_controller.go
booxter: Not sure what the value is though. Even for ovnController handler (current state of the repo), it's not immediately clear why we generate only when attachments are present. (It doesn't seem to buy us anything?) I guess it's ok either way, I just don't see the need to introduce additional check when the worst that can happen is an unused ConfigMap. But I won't insist.

controllers/ovndbcluster_controller.go
booxter: OK I see it was suggested in another thread. I commented there: https://github.com/openstack-k8s-operators/ovn-operator/pull/352#discussion_r1783177268 but I won't insist on a change here. :)

controllers/ovncontroller_controller.go
averdagu: I think so, I didn't want to modify it as I thought it was outside of the scope of the fix (as in the current code if this happens it won't return an error). But I agree that this is a configuration (no SB enabled/spawned) that we don't support.

My only question here is, could it happen that we reach this point because the OVN resources (ONVController,OVNDBCluster,OVNNorthd) were just created, and OVNController was the first one to be reconciled and when this point is reached it's just that OVNDBCluster was "slower" and it's not yet created? If we return an error, OVNController will get reconciled again later? Or it will remain on error state?

controllers/ovndbcluster_controller.go
averdagu: Done

controllers/ovndbcluster_controller.go
averdagu: Oh, thanks! Done

controllers/ovndbcluster_controller.go
averdagu: I think you're correct here, I'm not familiar with most of the code of openstack-operator, hence I don't know who creates the [openstackdataplaneservices.dataplane.openstack.org/ovn](https://logserver.rdoproject.org/openstack-periodic-container-antelope-centos9/review.rdoproject.org/rdo-jobs/master/periodic-adoption-multinode-to-crc-ceph/97f28e3/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/crs/openstackdataplaneservices.dataplane.openstack.org/ovn.yaml) and where the `ovncontroller-config` comes from. But either we find another way to get the name correct or it's another reason to split the config maps used (or remove the ovnEncapType as mentioned earlier)

averdagu: So, I asked on slack and currently we're creating this config map the wrong way, as the name of the config map used by edpm deployment is also a magic string. Openstack-operator is using this [file to create the OpenStackDataPlaneService](https://github.com/openstack-k8s-operators/openstack-operator/blob/main/config/services/dataplane_v1beta1_openstackdataplaneservice_ovn.yaml).

tests/functional/ovndbcluster_controller_test.go
averdagu: Regarding creating a test case with OVNController named differently to ovncontroller, currently the tests creates an OVNController object called "name := "ovncontroller-" + uuid.New().String()"

api/v1beta1/client.go
```diff
@@ -87,6 +107,37 @@ func getItems(list client.ObjectList) []client.Object {
 	return items
 }
 
+// OVNControllerNamespaceMapFunc - OVNController Watch Function
+func OVNControllerNamespaceMapFunc(crs client.ObjectList, reader client.Reader) handler.MapFunc {
```
booxter, line 111: why do you have to copy it instead of reusing the existing one? (Maybe rename it.)

controllers/ovncontroller_controller.go
booxter: I've thought about it for a bit, and I think now that the current logic is fine. It will not do anything until SB is created. When it's created, the reconcile for OVNController will trigger again and it will have a chance to handle it. Until then, from the perspective of the OVNController, there is no work to do. It's a success.

About your ordering concern, I think the above addresses it: when SB is up, OVNController will have another change to reconcile.

controllers/ovndbcluster_controller.go
```diff
@@ -86,6 +87,7 @@ func (r *OVNDBClusterReconciler) GetLogger(ctx context.Context) logr.Logger {
 }
 
 //+kubebuilder:rbac:groups=ovn.openstack.org,resources=ovndbclusters,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=ovn.openstack.org,resources=ovncontroller,verbs=create;update;patch;delete
```
booxter, line 90: do we really need to create/update/delete?

controllers/ovndbcluster_controller.go
```diff
@@ -612,6 +625,16 @@ func (r *OVNDBClusterReconciler) reconcileNormal(ctx context.Context, instance *
 
 		// Set DB Address
 		instance.Status.InternalDBAddress = strings.Join(internalDbAddress, ",")
+		if instance.Spec.DBType == ovnv1.SBDBType && instance.Spec.NetworkAttachment != "" {
```
booxter, line 628: what's the reason this is embedded under ReadyCount check? I think we should create the ConfigMap regardless of the current state of db pods.

booxter (changes requested):
Some minor things, I think the next respin will be ready for merge (IMO).

Please also add an explanation to the commit message about the change in how the name for -config map is constructed (now it's hardcoded), e.g. why we believe it's the right thing to do. (we should capture these details for posterity.)

controllers/ovndbcluster_controller.go
averdagu: Yeah.
Creation is needed bc we need to add the OVNEncapType once OVNController recreated
Update bc it could happen that user modifies the ExternalIDs.EncapType and configMap needs to update value
Delete is needed bc we want to remove OVNEncapType from config map

karelyatin: But these rules are for ovncontroller not configmap, and ovndbcluster controller don't need to create/delete/update ovncontrollers.

controllers/ovndbcluster_controller.go
averdagu: It was done to try to mimic what OVNController was doing. Also it was to ensure that when the configMap was created, environment was already ready to be used by EDPM nodes.
Also, given the time difference between creating configMap earlier on the reconcile loop or here would/should be minimal (order of units of seconds?) which is negligible compared with the EDPM deployment time. I think it's better to create configMap once pods are ready in order to catch the error earlier (EDPM deployment failed due to configMap not existing vs edpm not working correctly bc can't connect to SB).
Let me know what you think

controllers/ovndbcluster_controller.go
averdagu: Oh, true, I miss understood the verbs, changing it, thanks for noticing guys :)

api/v1beta1/client.go
averdagu: Yeah, will try to rename it so OVNCluster and OVNDBCluster uses same genericFunction

averdagu:
I hope I addressed all comments, I tested on a devnest and it worked fine. Config map is created regardless of OVNController being present, on update of OVNController, config map gets updated, as well as on deletion or creation of OVNController CR. So far I was not able to see ovn-operator stuck in an endless loop of reconcile

controllers/ovndbcluster_controller.go
booxter: I kinda disagree (one should be able to deploy EDP regardless of the current state of the SB service - otherwise a hiccup in SB may break the EDP deployment; we may want to make sure that ovn-controller on EDP starts successfully even without SB, but I think it should already do that - it should gracefully reconnect.)

I don't insist.

booxter (approved):

PR #357, opened 2024-09-27
Reset connection when turning off TLS

olliewalsh:
Jira: [OSPRH-10415](https://issues.redhat.com//browse/OSPRH-10415)


booxter:
The code looks correct. Should we add a kuttl test to validate that tls disabled removes the records? (Not sure it affects more than that. Does the service still start successfully if these records are left lingering?)

olliewalsh:
> The code looks correct. Should we add a kuttl test to validate that tls disabled removes the records? (Not sure it affects more than that. Does the service still start successfully if these records are left lingering?)

Basic tests work for me at least -  Disabling podLevel TLS on the controlplane and dataplane succeeds, previously neutronapi would fail to connect the the sb db.

PR #360, opened 2024-10-09
Add Liveness and Readiness probes to ovn-controller

fernandoroyosanchez:
Implement liveness and readiness probes to monitor the health of the ovn-controller.

Resolves: [OSPRH-6673](https://issues.redhat.com//browse/OSPRH-6673)

### DIFF
--- a/pkg/ovncontroller/daemonset.go
+++ b/pkg/ovncontroller/daemonset.go
@@ -62,6 +62,32 @@ func CreateOVNDaemonSet(
 		}...)
 	}
 
+	ovnControllerLivenessProbe := &corev1.Probe{
+		// TODO might need tuning
+		TimeoutSeconds:      5,
+		PeriodSeconds:       5,
+		InitialDelaySeconds: 30,
+	}
+
+	ovnControllerReadinessProbe := &corev1.Probe{
+		// TODO might need tuning
+		TimeoutSeconds:      5,
+		PeriodSeconds:       5,
+		InitialDelaySeconds: 30,
+	}
+
+	ovnControllerLivenessProbe.Exec = &corev1.ExecAction{
+		Command: []string{
+			"/usr/local/bin/container-scripts/ovn_controller_liveness.sh",
+		},
+	}
+
+	ovnControllerReadinessProbe.Exec = &corev1.ExecAction{
+		Command: []string{
+			"/usr/local/bin/container-scripts/ovn_controller_readiness.sh",
+		},
+	}
+
 	runAsUser := int64(0)
 	privileged := true
 
@@ -88,8 +114,10 @@ func CreateOVNDaemonSet(
 				RunAsUser:  &runAsUser,
 				Privileged: &privileged,
 			},
-			Env:          env.MergeEnvs([]corev1.EnvVar{}, envVars),
-			VolumeMounts: mounts,
+			Env:            env.MergeEnvs([]corev1.EnvVar{}, envVars),
+			VolumeMounts:   mounts,
+			ReadinessProbe: ovnControllerReadinessProbe,
+			LivenessProbe:  ovnControllerLivenessProbe,
 			// TODO: consider the fact that resources are now double booked
 			Resources:                instance.Spec.Resources,
 			TerminationMessagePolicy: corev1.TerminationMessageFallbackToLogsOnError,
@@ -148,16 +176,39 @@ func CreateOVSDaemonSet(
 		InitialDelaySeconds: 3,
 	}
 
+	ovsDbReadinessProbe := &corev1.Probe{
+		// TODO might need tuning
+		TimeoutSeconds:      5,
+		PeriodSeconds:       3,
+		InitialDelaySeconds: 3,
+	}
+
+	ovsVswitchdReadinessProbe := &corev1.Probe{
+		// TODO might need tuning
+		TimeoutSeconds:      5,
+		PeriodSeconds:       3,
+		InitialDelaySeconds: 3,
+	}
+
 	ovsDbLivenessProbe.Exec = &corev1.ExecAction{
 		Command: []string{
-			"/usr/bin/ovs-vsctl",
-			"show",
+			"/usr/local/bin/container-scripts/ovsdb_server_liveness.sh",
 		},
 	}
 	ovsVswitchdLivenessProbe.Exec = &corev1.ExecAction{
 		Command: []string{
-			"/usr/bin/ovs-appctl",
-			"bond/show",
+			"/usr/local/bin/container-scripts/vswitchd_liveness.sh",
+		},
+	}
+
+	ovsDbReadinessProbe.Exec = &corev1.ExecAction{
+		Command: []string{
+			"/usr/local/bin/container-scripts/ovsdb_server_readiness.sh",
+		},
+	}
+	ovsVswitchdReadinessProbe.Exec = &corev1.ExecAction{
+		Command: []string{
+			"/usr/local/bin/container-scripts/vswitchd_readiness.sh",
 		},
 	}
 
@@ -211,6 +262,7 @@ func CreateOVSDaemonSet(
 			// TODO: consider the fact that resources are now double booked
 			Resources:                instance.Spec.Resources,
 			LivenessProbe:            ovsDbLivenessProbe,
+			ReadinessProbe:           ovsDbReadinessProbe,
 			TerminationMessagePolicy: corev1.TerminationMessageFallbackToLogsOnError,
 		},
 		{
@@ -237,6 +289,7 @@ func CreateOVSDaemonSet(
 			// TODO: consider the fact that resources are now double booked
 			Resources:                instance.Spec.Resources,
 			LivenessProbe:            ovsVswitchdLivenessProbe,
+			ReadinessProbe:           ovsVswitchdReadinessProbe,
 			TerminationMessagePolicy: corev1.TerminationMessageFallbackToLogsOnError,
 		},
 	}
--- a/templates/ovncontroller/bin/ovsdb_server_liveness.sh
+++ b/templates/ovncontroller/bin/ovsdb_server_liveness.sh
@@ -1,0 +1,26 @@
+#!/bin/bash
+
+set -e
+
+error_exit() {
+    echo "$1" >&2
+    exit 1
+}
+
+# Check if ovsdb-server is running
+check_ovsdb_server_pid() {
+    if ! pidof -q ovsdb-server; then
+        error_exit "ERROR - ovsdb-server is not running"
+    fi
+}
+
+# Function to check the running status of ovsdb-server
+check_ovsdb_server_status() {
+    if ! /usr/bin/ovs-vsctl show 2>&1; then
+        error_exit "ERROR - Failed to get output from ovsdb-server, ovs-vsctl exit status: $?"
+    fi
+}
+
+
+check_ovsdb_server_pid
+check_ovsdb_server_status
--- a/templates/ovncontroller/bin/vswitchd_liveness.sh
+++ b/templates/ovncontroller/bin/vswitchd_liveness.sh
@@ -1,0 +1,26 @@
+#!/bin/bash
+
+set -e
+
+error_exit() {
+    echo "$1" >&2
+    exit 1
+}
+
+# Check if ovs-vswitchd is running
+check_ovs_vswitchd_pid() {
+    if ! pidof -q ovs-vswitchd; then
+        error_exit "ERROR - ovs-vswitchd is not running"
+    fi
+}
+
+# Function to check the running status of ovs-vswitchd
+check_ovs_vswitchd_status() {
+    if ! /usr/bin/ovs-appctl bond/show 2>&1; then
+        error_exit "ERROR - Failed to get output from ovs-vswitchd, ovs-appctl exit status: $?"
+    fi
+}
+
+
+check_ovs_vswitchd_pid
+check_ovs_vswitchd_status
--- a/tests/kuttl/tests/ovn_liveness/02-assert.yaml
+++ b/tests/kuttl/tests/ovn_liveness/02-assert.yaml
@@ -1,0 +1,14 @@
+#
+# Check for:
+#
+# - LivenessProbe failure in ovn-controller should restart it once
+#
+apiVersion: v1
+kind: Pod
+metadata:
+  labels:
+    service: ovn-controller
+status:
+  containerStatuses:
+    - name: ovn-controller
+      restartCount: 1
--- a/tests/kuttl/tests/ovn_liveness/02-pause-ovn-controller.yaml
+++ b/tests/kuttl/tests/ovn_liveness/02-pause-ovn-controller.yaml
@@ -1,0 +1,12 @@
+#
+# Step for:
+#
+# - Force one of the liveness checks to return fail in order to restart
+# ovn-controller
+#
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+commands:
+  - script: |
+      controller_pod=$(oc get pod -n $NAMESPACE -l service=ovn-controller -o name|head -1)
+      oc rsh -n $NAMESPACE ${controller_pod} ovn-appctl -t ovn-controller debug/pause
--- a/templates/ovncontroller/bin/ovn_controller_liveness.sh
+++ b/templates/ovncontroller/bin/ovn_controller_liveness.sh
@@ -1,0 +1,31 @@
+#!/bin/bash
+
+set -e
+
+error_exit() {
+    echo "$1" >&2
+    exit 1
+}
+
+# Check if ovn-controller is running
+check_ovn_controller_pid() {
+    if ! pidof -q ovn-controller; then
+        error_exit "ERROR - ovn-controller is not running"
+    fi
+}
+
+# Function to check the running status of ovn-controller
+check_ovn_controller_status() {
+    # Capture output and check exit status
+    if ! output=$(ovn-appctl -t ovn-controller debug/status 2>&1); then
+        error_exit "ERROR - Failed to get status from ovn-controller, ovn-appctl exit status: $?"
+    fi
+
+    if [ "$output" != "running" ]; then
+        error_exit "ERROR - ovn-controller status is '$output', expecting 'running' status"
+    fi
+}
+
+
+check_ovn_controller_pid
+check_ovn_controller_status
--- a/templates/ovncontroller/bin/ovn_controller_readiness.sh
+++ b/templates/ovncontroller/bin/ovn_controller_readiness.sh
@@ -1,0 +1,22 @@
+#!/bin/bash
+
+set -e
+
+error_exit() {
+    echo "$1" >&2
+    exit 1
+}
+
+# Check if ovn-controller is connected to the OVN SB database
+check_ovn_controller_connection() {
+    if ! output=$(ovn-appctl -t ovn-controller connection-status 2>&1); then
+        error_exit "ERROR - Failed to get connection status from ovn-controller, ovn-appctl exit status: $?"
+    fi
+
+    if [ "$output" != "connected" ]; then
+        error_exit "ERROR - ovn-controller connection status is '$output', expecting 'connected' status"
+    fi
+}
+
+
+check_ovn_controller_connection
--- a/templates/ovncontroller/bin/ovsdb_server_readiness.sh
+++ b/templates/ovncontroller/bin/ovsdb_server_readiness.sh
@@ -1,0 +1,26 @@
+#!/bin/bash
+
+set -e
+
+error_exit() {
+    echo "$1" >&2
+    exit 1
+}
+
+# Check ovsdb-server status
+check_ovsdb_server_status() {
+
+    if ! pid=$(cat /var/run/openvswitch/ovsdb-server.pid); then
+        error_exit "ERROR - Failed to get pid for ovsdb-server, exit status: $?"
+    fi
+
+    if ! output=$(ovs-appctl -t /var/run/openvswitch/ovsdb-server."$pid".ctl ovsdb-server/list-dbs); then
+        error_exit "ERROR - Failed retrieving list of databases from ovsdb-server"
+    fi
+
+    if [[ "$output" != *"Open_vSwitch"* ]]; then
+        error_exit "ERROR - 'Open_vSwitch' not found on databases"
+    fi
+}
+
+check_ovsdb_server_status
--- a/templates/ovncontroller/bin/vswitchd_readiness.sh
+++ b/templates/ovncontroller/bin/vswitchd_readiness.sh
@@ -1,0 +1,26 @@
+#!/bin/bash
+
+set -e
+
+error_exit() {
+    echo "$1" >&2
+    exit 1
+}
+
+# Check ovs-vswitchd status
+check_ovs_vswitchd_status() {
+
+    if ! pid=$(cat /var/run/openvswitch/ovs-vswitchd.pid); then
+        error_exit "ERROR - Failed to get pid for ovs-vswitchd, exit status: $?"
+    fi
+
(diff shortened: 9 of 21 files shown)


mtomaska (changes requested):

templates/ovncontroller/bin/liveness.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/sh
+
+# Ensure ovn-controller pid is up
+output=$(pidof ovn-controller)
+if [ "$output" != "1" ]; then
```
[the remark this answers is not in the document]
fernandoroyosanchez: I think so, command specified in the container is :

```
     Command:                                                                                                                                                                                                                                                                                                               
│       ovn-controller                                                                                                                                                                                                                                                                                                     
│       --pidfile                                                                                                                                                                                                                                                                                                            
│       unix:/run/openvswitch/db.sock                                                                                                                                                                                                                                                                                        
│       --certificate=/etc/pki/tls/certs/ovndb.crt                                                                                                                                                                                                                                                                           
│       --private-key=/etc/pki/tls/private/ovndb.key                                                                                                                                                                                                                                                                        
│       --ca-cert=/etc/pki/tls/certs/ovndbca.crt  
```

IIUC, in Kubernetes, the command executed in the container will be the process with PID 1 within the container's namespace.

templates/ovncontroller/bin/liveness.sh
```diff
@@ -0,0 +1,18 @@
...
+    echo "ERROR - OVN-Controller is not running"
+    exit 1
+fi
+
+# Ensure ovn-controller is running and not paused, this running status is not
+# affected by any other services/depedencies as the connection-status output
+output=$(ovn-appctl -t ovn-controller debug/status)
+if [ "$output" != "running" ]; then
```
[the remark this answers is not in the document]
fernandoroyosanchez: After taking a loop to that test file, looks like the prefix x is commonly used in shell scripts to avoid issues with empty strings. If the command output is an empty string, then x$(...) will simply be x, which prevents comparison errors when there is no concrete output, so let me do that kind of comparison on the code.

templates/ovncontroller/bin/readiness.sh
```diff
@@ -0,0 +1,38 @@
...
+    echo "ERROR - Failed to get pid for ovsdb-server process"
+    exit 1
+fi
+
+# Ensure ovs-vswitchd is OK
+ovsvswitchdpidfile=$(find /var/run/openvswitch -name "ovs-vswitchd.pid")
+if [ -n "$ovsvswitchdpidfile" ]; then
+    pid=$(cat "$ovsvswitchdpidfile")
+    output=$(ovn-appctl -t /var/run/openvswitch/ovs-vswitchd."$pid".ctl ofproto/list)
+    if [ -z "$output" ]; then
```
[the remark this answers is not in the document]
fernandoroyosanchez: Let me apply same commented on previous item to apply here.

templates/ovncontroller/bin/liveness.sh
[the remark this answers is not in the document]
fernandoroyosanchez: Finally checking debug/status command is getting normal output and no issues where there running the command, if it pass then checking "running" output

templates/ovncontroller/bin/readiness.sh
```diff
@@ -0,0 +1,73 @@
...
+        error_exit "ERROR - Failed to get connection status from ovn-controller"
+    fi
+    if [ "$output" != "connected" ]; then
+        error_exit "ERROR - OVN-Controller is not connected to OVN SB database"
+    fi
+}
+
+# Check if ovsdb-server is running and list its databases
+check_ovsdb_server() {
+    ovsdbserverpidfile=$(find /var/run/openvswitch -name "ovsdb-server.pid")
```
booxter, line 22: do we really need to call `find` to find the file? Can't it be a hardcoded name since we probably already know the exact location?

templates/ovncontroller/bin/readiness.sh
```diff
@@ -0,0 +1,73 @@
...
+    if [ $? -ne 0 ]; then
+        error_exit "ERROR - Failed to get connection status from ovn-controller"
+    fi
+    if [ "$output" != "connected" ]; then
+        error_exit "ERROR - OVN-Controller is not connected to OVN SB database"
+    fi
+}
+
+# Check if ovsdb-server is running and list its databases
+check_ovsdb_server() {
```
booxter, line 21: Question:

Shouldn't this check belong to where the ovsdb-server actually runs - to the `-ovs` pod checks? From `ovn-controller` pod perspective, you already confirmed that `connection-status` is fine for SB. I don't think there's an interface to check that its vswitchd openflow socket and ovsdb-server local connection is established. Maybe OVN should add commands for these?

templates/ovncontroller/bin/readiness.sh
```diff
@@ -0,0 +1,73 @@
...
+    if [ -z "$pid" ]; then
+        error_exit "ERROR - Failed to read PID from ovsdb-server PID file"
+    fi
+
+    output=$(ovn-appctl -t /var/run/openvswitch/ovsdb-server."$pid".ctl ovsdb-server/list-dbs)
+    if [ $? -ne 0 ]; then
+        error_exit "ERROR - Failed to communicate with ovsdb-server"
+    fi
+
+    if [ -z "$output" ]; then
```
booxter, line 37: If you go with it, at least confirm that the expected database name is in the output? (Not sure we should do the check for any of these though - this strikes me as an implementation detail that the config is in a local ovsdb-server.)

booxter:
This is directionally correct. My concerns:

0. All new shell code should enable `set -e`, otherwise it will swallow any errors and we won't know about it.
1. I think some of the checks belong to `-ovs` pod, not `ovn-controller`.
2. I am not sure it's safe / needed to do vswitchd / ovsdb-server sanity checks in `ovn-controller`.

The `connection-status` and `debug/status` checks are good since they don't rely on any particular implementation but ask the service itself about what it believes its own health is. As long as the service claims health, we should believe it. What you probably try to address is the fact that e.g. `connection-status` doesn't check if openflow / local ovsdb-server connection is established? Perhaps we should report a feature request for this for FDP team and let them enhance their interface?

booxter (changes requested):

templates/ovncontroller/bin/liveness.sh
```diff
@@ -0,0 +1,40 @@
...
+    fi
+
+    if [ "$output" != "1" ]; then
+        error_exit "ERROR - OVN-Controller is not running"
+    fi
+}
+
+# Function to check the running status of ovn-controller
+check_ovn_controller_status() {
+    output=$(ovn-appctl -t ovn-controller debug/status 2>/dev/null)
```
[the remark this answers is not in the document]
fernandoroyosanchez: typo keep here, not suppresing stderr in any of the commands

templates/ovncontroller/bin/readiness.sh
```diff
@@ -0,0 +1,73 @@
...
+error_exit() {
+    echo "$1"
+    exit 1
+}
+
+# Check if ovn-controller is connected to the OVN SB database
+check_ovn_controller_connection() {
+    output=$(ovn-appctl -t ovn-controller connection-status)
+    if [ $? -ne 0 ]; then
+        error_exit "ERROR - Failed to get connection status from ovn-controller"
```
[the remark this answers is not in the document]
fernandoroyosanchez: make sense

templates/ovncontroller/bin/liveness.sh
```diff
@@ -0,0 +1,40 @@
...
+
+# Function to handle errors and exit with a message
+error_exit() {
+    echo "$1"
+    exit 1
+}
+
+# Check if ovn-controller is running
+check_ovn_controller_pid() {
+    output=$(pidof ovn-controller)
```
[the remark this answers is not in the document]
fernandoroyosanchez: good one! done!

templates/ovncontroller/bin/readiness.sh
fernandoroyosanchez: done, after checking that file is same path on dev env and a real one.

templates/ovncontroller/bin/readiness.sh
fernandoroyosanchez: By previous comment reason, removing last checks over ovsdb-server and ovs-vswitchd.

templates/ovncontroller/bin/readiness.sh
fernandoroyosanchez: Yes, I understand your comment that the last two checks for ovsdb-server and ovs-vswitchd are directly related to the ready status of those services. For ovn-controller, it should be sufficient to confirm that the connection-status = 'connected', indicating it's connected to the database.

Considering discussions with other stakeholders, such as the telemetry team, who are concerned about the time impact of running health checks every second, we should aim to keep these checks as simple as possible. Each service should only run the necessary checks, and we should trust that other services will do the same. Everyone is responsible for ensuring their own checks run in a timely manner.

So removing last checks over ovsdb-server and ovs-vswitchd.

fernandoroyosanchez:
Thx for your review!!

> This is directionally correct. My concerns:
> 
> 0. All new shell code should enable `set -e`, otherwise it will swallow any errors and we won't know about it.

ack

> 1. I think some of the checks belong to `-ovs` pod, not `ovn-controller`.
> 2. I am not sure it's safe / needed to do vswitchd / ovsdb-server sanity checks in `ovn-controller`.

I put them in the readiness checks just because they are compromising the ovn-controller ready to manage events from OVN SB DB, indeed readiness should not restart the container due to those checks. 

> 3. Tests should be added to validate the checks don't misbehave.

Sure, I will add!

> 
> The `connection-status` and `debug/status` checks are good since they don't rely on any particular implementation but ask the service itself about what it believes its own health is.  As long as the service claims health, we should believe it. What you probably try to address is the fact that e.g. `connection-status` doesn't check if openflow / local ovsdb-server  connection is established? 

`connection-status`  has dependencies with the ovsdbserver-sb, (if i'm not wrong), it shows OVN SBDB connection status for the chassis, so only `debug/status` was considered for liveness after confirming that point with ovn folks.

> Perhaps we should report a feature request for this for FDP team and let them enhance their interface?

Yes, they confirm that debugging / status is not documented, but we can use it, but yes an improvement there would pave our way.

templates/ovncontroller/bin/readiness.sh
```diff
@@ -0,0 +1,53 @@
...
+        error_exit "ERROR - ovsdb-server PID file does not exist"
+    fi
+
+    pid=$(cat "$ovsdbserverpidfile")
+    if [ -z "$pid" ]; then
+        error_exit "ERROR - Failed to read PID from ovsdb-server PID file"
+    fi
+}
+
+# Check if ovs-vswitchd is running and list ofproto instances
```
karelyatin, line 34: comment looks outdated as mentions ofproto instances.

templates/ovncontroller/bin/readiness.sh
```diff
@@ -0,0 +1,53 @@
...
+    output=$(ovn-appctl -t ovn-controller connection-status)
+    if [ $? -ne 0 ]; then
+        error_exit "ERROR - Failed to get connection status from ovn-controller, ovn-appctl exit status: $?"
+    fi
+    if [ "$output" != "connected" ]; then
+        error_exit "ERROR - ovn-controller connection status is '$output', expecting 'connected' status"
+    fi
+}
+
+# Check if ovsdb-server is running and list its databases
```
karelyatin, line 21: this too oudated as there is no list db below

pkg/ovncontroller/daemonset.go
```diff
@@ -62,6 +62,32 @@ func CreateOVNDaemonSet(
...
+		TimeoutSeconds:      5,
+		PeriodSeconds:       5,
+		InitialDelaySeconds: 30,
+	}
+
+	ovnControllerReadinessProbe := &corev1.Probe{
+		// TODO might need tuning
+		TimeoutSeconds:      5,
+		PeriodSeconds:       5,
+		InitialDelaySeconds: 10,
```
karelyatin, line 76: Noticing the difference in Iniitaldelaysecond curious on how these were computed? normally ready time > live time but here it's opposite.

templates/ovncontroller/bin/liveness.sh
```diff
@@ -0,0 +1,34 @@
...
+}
+
+# Function to check the running status of ovn-controller
+check_ovn_controller_status() {
+    output=$(ovn-appctl -t ovn-controller debug/status)
+    if [ $? -ne 0 ]; then
+        error_exit "ERROR - Failed to get status from ovn-controller, ovn-appctl exit status: $?"
+    fi
+    if [ "$output" != "running" ]; then
+        error_exit "ERROR - ovn-controller is reporting a $output status, instead of expected running status"
```
karelyatin, line 25: can be consistent in formatting same as readiness probe

templates/ovncontroller/bin/readiness.sh
```diff
@@ -0,0 +1,53 @@
...
+error_exit() {
+    echo "$1"
+    exit 1
+}
+
+# Check if ovn-controller is connected to the OVN SB database
+check_ovn_controller_connection() {
+    output=$(ovn-appctl -t ovn-controller connection-status)
+    if [ $? -ne 0 ]; then
+        error_exit "ERROR - Failed to get connection status from ovn-controller, ovn-appctl exit status: $?"
```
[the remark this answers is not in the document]
fernandoroyosanchez: Our debug messages will not appear, but the return code will be != 0. For example, a fake event generated intentionally will return an exit code of 2. Obviously, we won't have our subsequent debug messages, but we do have the command output.

`Warning  Unhealthy       1s (x16 over 71s)  kubelet            Readiness probe failed: ls: cannot access '/home/fake': No such file or directory`

pkg/ovncontroller/daemonset.go
fernandoroyosanchez: These were initial values, as noted in the TODO, and they may need tuning. The reason I set the livenessProbe to 30s is to ensure that the container has enough time to start before the liveness check can potentially cause a restart of the pod. The readinessProbe is less critical, as it would, at worst, leave the container unhealthy. However, upon further reflection, I don’t see the sense in having readiness checks before the first liveness check, so I’ll set it to the same value! We will have time to tune them in a future PR if necessary.

templates/ovncontroller/bin/readiness.sh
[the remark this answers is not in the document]
karelyatin: still not clear if the debug message not logged what's the role of these? with set -e the script immediately terminates on any kind of error.

tests/kuttl/tests/ovn_liveness/02-assert.yaml
```diff
@@ -0,0 +1,14 @@
+#
+# Check for:
+#
+# - LivenessProbe in ovn-controller should be restart it once
```
karelyatin, line 4: LivenessProbe failure in ovn-controller should restart it once

tests/kuttl/tests/ovn_readiness/02-patch.yaml
```diff
@@ -0,0 +1,13 @@
+#
+# Step for:
+#
+# - Reduce the failureThreshold and periodSeconds to detect a failure in OVN SB DB before it reconciles
+#
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+commands:
+    - script: |
+        oc patch daemonset ovn-controller -n $NAMESPACE --type='json' -p='[
```
karelyatin, line 10: any changes to daemonset will not persist, but could trigger container restart, but seems may have issues some time.
what about doing it differently and avoid this steps and sb pod delete and ready false checks below?
- check connection status and it should be connected
- configure ovn-remote to some invalid address
- check connection-status and it should be not != connected

templates/ovncontroller/bin/readiness.sh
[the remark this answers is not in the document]
fernandoroyosanchez: Finally removing set -e as explicit 0 or 1 set.

templates/ovncontroller/bin/liveness.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/sh
```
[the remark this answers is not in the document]
fernandoroyosanchez: Finally, we removed set -e to capture a specific error message when $? is not equal to 0. This way, the output includes not only our custom error message, but also the output from the failed command. Let me know if you consider any further option.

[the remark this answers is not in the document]
booxter: This is because you should not check `$?` but use `! cmd` syntax. See:

```
$ cat /tmp/test.sh
#!/bin/bash

set -e

error_exit() {
	echo "Logging!"
	exit 1
}

if ! false; then
	error_exit
fi

$ cat /tmp/test2.sh
#!/bin/bash

set -e

error_exit() {
	echo "Logging!"
	exit 1
}

false
if [ $? -ne 0 ]; then
	error_exit
fi

$ bash /tmp/test.sh
Logging!

$ bash /tmp/test2.sh
$
```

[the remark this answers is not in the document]
booxter: This is documented in the official bash docs. See: https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html

```
The shell does not exit if the command that fails is part of the command list immediately following a while or until keyword, part of the test in an if statement, part of any command executed in a && or || list except the command following the final && or ||, any command in a pipeline but the last, or if the command’s return status is being inverted with !.
```

templates/ovncontroller/bin/readiness.sh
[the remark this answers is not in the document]
booxter: I'd rather have `set -e` enabled and deal with particular situations where the return code should be ignored, e.g. with `output=$(ovn-appctl ... || true)`.

templates/ovncontroller/bin/readiness.sh
booxter: Have you removed the checks? I still see them @fernandoroyosanchez

templates/ovncontroller/bin/readiness.sh
```diff
@@ -0,0 +1,50 @@
...
+        error_exit "ERROR - Failed to read PID from ovs-vswitchd PID file"
+    fi
+}
+
+
+check_ovn_controller_connection
+check_ovsdb_server
+check_ovs_vswitchd
+
+exit 0
```
booxter, line 50: nit: redundant

tests/kuttl/tests/ovn_liveness/02-pause-ovn-controller.yaml
```diff
@@ -0,0 +1,12 @@
...
+#
+# - Force one of the liveness checks to return fail in order to restart
+# ovn-controller
+#
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+commands:
+  - script: |
+      controller_pod=$(oc get pod -n $NAMESPACE -l service=ovn-controller -o name|head -1)
+      oc rsh -n $NAMESPACE ${controller_pod} ovn-appctl -t ovn-controller debug/pause
```
booxter, line 12: i like it

booxter (changes requested):
1. I still think we should have a policy of always using `set -e` for defensive programming. There are ways to handle it for special sections of code that have to ignore result.
2. I'm not sure why ovn-controller pod still checks that -ovs pod is working. Shouldn't the -ovs pod do the checks for vswitchd / ovsdb-server?

templates/ovncontroller/bin/liveness.sh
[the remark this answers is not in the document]
booxter: I've noticed your script (claims to be) POSIX shell (`/bin/sh`), so I figured I should check [POSIX standard](https://pubs.opengroup.org/onlinepubs/007904875/utilities/set.html).

```
-e
    When this option is on, if a simple command fails for any of the reasons listed in [Consequences of Shell Errors](https://pubs.opengroup.org/onlinepubs/007904875/utilities/xcu_chap02.html#tag_02_08_01) or returns an exit status value >0, and is not part of the compound list following a while, until, or if keyword, and is not a part of an AND or OR list, and is not a pipeline preceded by the ! reserved word, then the shell shall immediately exit.
```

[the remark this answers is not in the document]
fernandoroyosanchez: Switching to bash since the simplicity of the script allows it, and it seems that certain flags can cause issues (e.g. pidof -q O_o) after reading some posts. Didn't know the `! cmd` and looks very clean, so I don’t have to keep checking `$?`

templates/ovncontroller/bin/readiness.sh
fernandoroyosanchez: It makes sense now that I understood the ticket referred to creating liveness/readiness probes for all the containers and not just for ovn-controller. Thanks for insisting!

templates/ovncontroller/bin/readiness.sh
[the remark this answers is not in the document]
fernandoroyosanchez: I think last version addresses it

fernandoroyosanchez:
> 1. I still think we should have a policy of always using `set -e` for defensive programming. There are ways to handle it for special sections of code that have to ignore result.

`set -e` recover after apply your proposal `! cmd`. thx for that!

> 2. I'm not sure why ovn-controller pod still checks that -ovs pod is working. Shouldn't the -ovs pod do the checks for vswitchd / ovsdb-server?

After understanding the scope of the ticket, I separated the checks for the different containers, aiming to manage responsibilities independently. What I haven't been able to do yet is simulate a readiness failure for ovsdb-server or vswitch, as the "tricks" I'm trying in kuttl don't seem to work.

fernandoroyosanchez:
recheck

templates/ovncontroller/bin/vswitchd_liveness.sh
```diff
@@ -0,0 +1,26 @@
...
+
+error_exit() {
+    echo "$1" >&2
+    exit 1
+}
+
+# Check if ovs_vswitchd is running
+check_ovs_vswitchd_pid() {
+    if ! pidof -q ovs-vswitchd; then
+        error_exit "ERROR - ovs_vswitchd is not running"
```
karelyatin, line 13: ovs-vswitchd

templates/ovncontroller/bin/vswitchd_liveness.sh
```diff
@@ -0,0 +1,26 @@
...
+    fi
+}
+
+# Function to check the running status of ovs_vswitchd
+check_ovs_vswitchd_status() {
+    if ! /usr/bin/ovs-appctl bond/show 2>&1; then
+        error_exit "ERROR - Failed to get output from ovs_vswitchd, ovs-appctl exit status: $?"
```
karelyatin, line 20: ovs-vswitchd

templates/ovncontroller/bin/ovsdb_server_liveness.sh
```diff
@@ -0,0 +1,26 @@
...
+check_ovsdb_server_pid() {
+    if ! pidof -q ovsdb-server; then
+        error_exit "ERROR - ovsdb-server is not running"
+    fi
+}
+
+# Function to check the running status of ovsdb-server
+check_ovsdb_server_status() {
+    if ! /usr/bin/ovs-vsctl show 2>&1; then
+        error_exit "ERROR - Failed to get output from ovsdb_server, ovs-vsctl exit status: $?"
```
karelyatin, line 20: ovsdb-server

karelyatin (commented):
/lgtm

booxter (approved):

booxter:
```
 error: build error: Failed to push image: trying to reuse blob sha256:d2e5d6c907c71aea278da1e0e3424927afce03a253cba174efe5392fcba72b55 at destination: pinging container registry quay.rdoproject.org: Get "https://quay.rdoproject.org/v2/": dial tcp 38.129.56.158:443: i/o timeout 
```

PR #361, opened 2024-10-10
Force recreation of ovs db cluster on first replica when protocol changes

olliewalsh:
Cannot remove final cluster member to change protocol, but can convert to standalone schema. Cluster gets recreated with the correct address.

Jira: [OSPRH-6954](https://issues.redhat.com//browse/OSPRH-6954)
Jira: [OSPRH-10415](https://issues.redhat.com//browse/OSPRH-10415)

### DIFF
--- a/kuttl-test.yaml
+++ b/kuttl-test.yaml
@@ -19,7 +19,7 @@ kind: TestSuite
 reportFormat: JSON
 reportName: kuttl-test-ovn
 namespace: openstack
-timeout: 360
+timeout: 600
 parallel: 1
 suppress:
   - events                     # Remove spammy event logs
--- a/templates/ovndbcluster/bin/setup.sh
+++ b/templates/ovndbcluster/bin/setup.sh
@@ -87,6 +87,26 @@ if ! [ -s $DB_FILE ]; then
     cleanup_db_file
 fi
 
+# Must remove a cluster member to change protocol, replicas 1/2 will have
+# left the cluster when terminating the pod, but cannot remove final member from
+# a cluster (replica 0).
+# Convert db to standalone mode on this member instead.
+# Cluster then gets recreated by ovnctl run_*b_ovsdb using the new local address.
+if [ "$(hostname)" == "{{ .SERVICE_NAME }}-0" ]; then
+    DB_LOCAL_ADDR={{ if .TLS }}ssl{{ else }}tcp{{ end }}:$(hostname).{{ .SERVICE_NAME }}.${NAMESPACE}.svc.cluster.local:${RAFT_PORT}
+    if [ -e ${DB_FILE} ] && \
+       ovsdb-tool db-is-clustered ${DB_FILE} && \
+       ACTUAL_DB_LOCAL_ADDR="$(ovsdb-tool db-local-address ${DB_FILE})" && \
+       [ "${ACTUAL_DB_LOCAL_ADDR}" != "${DB_LOCAL_ADDR}" ] \
+       ; \
+    then
+        rm -f "${DB_FILE%.db}_standalone.db"
+        if ovsdb-tool cluster-to-standalone "${DB_FILE%.db}_standalone.db" "${DB_FILE}"; then
+            mv -f "${DB_FILE%.db}_standalone.db" "${DB_FILE}"
+        fi
+    fi
+fi
+
 # Wait until the ovsdb-tool finishes.
 trap wait_for_ovsdb_tool EXIT
 
@@ -106,8 +126,10 @@ if [[ "$(hostname)" == "{{ .SERVICE_NAME }}-0" ]]; then
 
 {{- if .TLS }}
     ${CTLCMD} set-ssl {{.OVNDB_KEY_PATH}} {{.OVNDB_CERT_PATH}} {{.OVNDB_CACERT_PATH}}
-    ${CTLCMD} set-connection ${DB_SCHEME}:${DB_PORT}:${DB_ADDR}
+{{- else }}
+    ${CTLCMD} del-ssl
 {{- end }}
+    ${CTLCMD} set-connection ${DB_SCHEME}:${DB_PORT}:${DB_ADDR}
 
     # OVN does not support setting inactivity-probe through --remote cli arg so
     # we have to set it after database is up.
--- a/tests/kuttl/tests/ovn_tls_enable/01-assert.yaml
+++ b/tests/kuttl/tests/ovn_tls_enable/01-assert.yaml
@@ -1,0 +1,39 @@
+apiVersion: ovn.openstack.org/v1beta1
+kind: OVNDBCluster
+metadata:
+  name: ovndbcluster-nb-sample
+spec:
+  replicas: 3
+status:
+  readyCount: 3
+---
+apiVersion: ovn.openstack.org/v1beta1
+kind: OVNDBCluster
+metadata:
+  name: ovndbcluster-sb-sample
+spec:
+  replicas: 3
+status:
+  readyCount: 3
+---
+apiVersion: kuttl.dev/v1beta1
+kind: TestAssert
+commands:
+  # check the DB uri scheme is tcp
+  - script: |
+      template='{{.status.internalDbAddress}}{{"\n"}}'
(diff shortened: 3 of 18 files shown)


templates/ovndbcluster/bin/setup.sh
```diff
@@ -87,6 +87,23 @@ if ! [ -s $DB_FILE ]; then
...
+# Must remove a cluster member to change protocol, replicas 1/2 will have
+# left the cluster when terminating the pod, but cannot remove final member from
+# a cluster (replica 0).
+# Convert db to standalone mode on this member instead.
+# Cluster then gets recreated by ovnctl run_*b_ovsdb using the new local address.
+if [ "$(hostname)" == "{{ .SERVICE_NAME }}-0" ]; then
+    DB_LOCAL_ADDR={{ if .TLS }}ssl{{ else }}tcp{{ end }}:$(hostname).{{ .SERVICE_NAME }}.${NAMESPACE}.svc.cluster.local:${RAFT_PORT}
+    if [ -e ${DB_FILE} ] && \
+       ovsdb-tool db-is-clustered ${DB_FILE} && \
+       [ "$(ovsdb-tool db-local-address ${DB_FILE})" != "${DB_LOCAL_ADDR}" ] \
```
olliewalsh, line 99: TODO: handle this command failing

templates/ovndbcluster/bin/setup.sh
```diff
@@ -87,6 +87,23 @@ if ! [ -s $DB_FILE ]; then
...
+       ; \
+    then
+        mv -f ${DB_FILE} /tmp/ovn_db_tmp_standalone.db
```
olliewalsh, line 102: TODOs (input from Ihar):
try to make this atomic/re-entrant
do we need to consider disk usage when converting large dbs?
do we need to consider ready/liveness probes when converting large dbs?

olliewalsh: Converting a large sb db took ~12s on my nvme backed crc deployment.

The current liveness probe seems extremely trigger-happy though:
```
		TimeoutSeconds:      5,
		PeriodSeconds:       3,
		InitialDelaySeconds: 3,
```
I think it would be best to add a startup probe too.

kuttl-test.yaml
```diff
@@ -19,7 +19,7 @@ kind: TestSuite
 reportFormat: JSON
 reportName: kuttl-test-ovn
 namespace: openstack
-timeout: 360
+timeout: 600
```
olliewalsh, line 22: TODO: take a closer look at this. When replicas=3 pods do not appear to terminating gracefully in ovn_tls_enable (cleanup step) and ovn_db_delete (delete-pods step)

olliewalsh: Cluster cleanup logic was the problem

olliewalsh: It's not trivial to resolve the cluster cleanup issues on delete, and not directly related to this PR, so for now will just increase the kuttl timeout.

templates/ovndbcluster/bin/setup.sh
karelyatin: I think if we move conversion tasks to an init container we do not need to care about these probe timeouts

kuttl-test.yaml
karelyatin: what was the issue with cluster cleanup? from a previous run i see:-

`    logger.go:42: 22:50:55 | ovn_db_delete/4-delete-pods | running command: [sh -c oc delete pods -n $NAMESPACE -l service=ovsdbserver-nb
        ]
    logger.go:42: 22:50:56 | ovn_db_delete/4-delete-pods | pod "ovsdbserver-nb-0" deleted
    logger.go:42: 22:50:56 | ovn_db_delete/4-delete-pods | pod "ovsdbserver-nb-1" deleted
    logger.go:42: 22:50:56 | ovn_db_delete/4-delete-pods | pod "ovsdbserver-nb-2" deleted
    logger.go:42: 22:56:55 | ovn_db_delete/4-delete-pods | command failure, skipping 1 additional commands
    case.go:378: failed in step 4-delete-pods
    case.go:380: command "oc delete pods -n $NAMESPACE -l service=ovsdbserver-nb" exceeded 360 sec timeout, context deadline exceeded`

seems delete command just got stuck and with timeout increase it shouldn't help

The current timeout was set to 360s as gracefultimeout is set to 5 minutes.

tests/kuttl/tests/ovn_tls_enable/01-assert.yaml
```diff
@@ -0,0 +1,39 @@
...
+commands:
+  # check the DB uri scheme is tcp
+  - script: |
+      template='{{.status.internalDbAddress}}{{"\n"}}'
+      regex="tcp:.*"
+      dbUri=$(oc get -n $NAMESPACE OVNDBCluster ovndbcluster-sb-sample -o go-template="$template")
+      matches=$(echo "$dbUri" | sed -e "s?$regex??")
+      if [[ -n "$matches" ]]; then
+        exit 1
+      fi
```
stuggi, line 30: would this be the same as ?
```
oc get -n $NAMESPACE OVNDBCluster ovndbcluster-sb -o jsonpath={.status.internalDbAddress} | grep -q tcp
```

tests/kuttl/tests/ovn_tls_enable/01-assert.yaml
```diff
@@ -0,0 +1,39 @@
...
+  # Check ovn connect is ptcp
+  - script: |
+      sb_pod=$(oc get pod -n $NAMESPACE -l service=ovsdbserver-sb -o name|head -1)
+      oc rsh -n $NAMESPACE ${sb_pod} ovn-sbctl --no-leader-only  get-connection | grep -q ptcp
```
stuggi, line 34: can use `oc rsh` or `oc exec` on the deployment to not have to check for the pod I think

tests/kuttl/tests/ovn_tls_enable/01-assert.yaml
olliewalsh: Yes, would use `grep -q '^tcp:'`.

tests/kuttl/tests/ovn_tls_enable/01-assert.yaml
olliewalsh: yes, seems to work
```
$ oc rsh statefulset/ovsdbserver-sb hostname
ovsdbserver-sb-0
```

This and the other assert were taken from the existing kutt tests so maybe clean them all up in a follow up patch?

stuggi: works for me

kuttl-test.yaml
olliewalsh: issue is that when the statefulset (or all of the pods) are deleted then pod-0 will exit before pods 1/2 which enter an endless loop waiting to leave the cluster. They are eventually killed but it can take take slightly longer than the current kuttl timeout for the dbcluster delete in the tls_enable test cleanup

karelyatin: ack but that still doesn't explain why that much time is required with 5 minute graceful timeout, but as you mentioned over slack the issue was hit locally and logs are no longer available so couldn't see why this case hit. so test improvement can be checked later. As you suggested one option to use --grace-period=30 while kill faster

But yes with tls test it's doing more ops could easily hit currently configured 6 minute timeout so should be fine to increase for now.

templates/ovndbcluster/bin/setup.sh
karelyatin: Mention over slack with a quite large db it started in 12 seconds, so should be fine to avoid init container just for this for now.

karelyatin (approved):
/lgtm

templates/ovndbcluster/bin/setup.sh
booxter: (For posterity, listing my understanding of this problem) the liveness check is `pidof ovsdb-server` so probably only initialization is a potential issue, and this seems to be addressed. If the liveness check ever gets more complicated, (e.g. ovsdb-server has to reply to a status command), these may be revisited.

PR #368, opened 2024-10-23
Restart ovn-controller gracefully on PreStop

booxter:
Otherwise, it will clean up chassis and other relevant records and cause unnecessary router failovers.

--

This is not a complete solution: the operator will have to indicate to ovn-controller when it's actually ok to clean the chassis records up, with the file created in the container. (Note: I am not sure this is the best way to reflect this request into the pod. Perhaps the pod should instead talk to k8s where this information could be exposed. Another alternative is storing this information somewhere in OVSDB - maybe ovn-controller could even be expanded to read the flag from there and do the right thing?)

Related-Issue: [OSPRH-10821](https://issues.redhat.com//browse/OSPRH-10821)


booxter:
@karelyatin FYI obviously I'm not going to complete / work on this PR. If it makes more sense, feel free to fork it into your own work space and close this PR.

PR #373, opened 2024-11-04
Allow exposing ovn DB using a k8s service

stuggi:
Using multus to attach the ovn db pods to the networks that the edpm nodes can access it won't work for BGP. A loadbalancer service, like metallb, could be used in this case.

This introduces service override to the ovndbcluster, like we have for other services which use a LB service to expose them to the outside. Adding the AnnotationHostnameKey annotation the service hostname gets automatic registered in the DNS instance.

In general this could replace all the current implementation of attaching the ovn db pods using multus, creating the headless service and creating the dnsdata to register it in dns. For update/backward compatability this is kept as if if no service override is provided.

To expose the ovndbcluster-sb via a metallb loadbalancer would look like:

```
  ovn:
    enabled: true
    template:
      ...
      ovnDBCluster:
        ovndbcluster-nb:
	  ...
        ovndbcluster-sb:
          dbType: SB
	  ...
          override:
            service:
              metadata:
                annotations:
                  metallb.universe.tf/address-pool: internalapi
                  metallb.universe.tf/allow-shared-ip: internalapi
                  metallb.universe.tf/loadBalancerIPs: 172.17.0.80
              spec:
                type: LoadBalancer
      ovnNorthd:
        ...
```

Jira: [OSPRH-651](https://issues.redhat.com//browse/OSPRH-651)


controllers/ovndbcluster_controller.go
```diff
@@ -670,135 +670,207 @@ func (r *OVNDBClusterReconciler) reconcileServices(
...
+			// have complete information, return error to retrigger reconcile loop
+			// Returning here instead of at the beggining of the for is done to
+			// expose the already created pods to other services/dataplane nodes
+			if len(podList.Items) < int(*(instance.Spec.Replicas)) {
+				Log.Info(fmt.Sprintf("not all pods are yet created, number of expected pods: %v, current pods: %v", *(instance.Spec.Replicas), len(podList.Items)))
+				return ctrl.Result{RequeueAfter: 1 * time.Second}, nil
 			}
+		}
 
+	} else {
```
stuggi, line 805: in theory I think we could remove all the above and use lb service all the time, but as mentioned in the description left it for backward compatibility until we'd know how to deal with it.

controllers/ovndbcluster_controller.go
```diff
@@ -670,135 +670,207 @@ func (r *OVNDBClusterReconciler) reconcileServices(
...
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      serviceName,
+				Namespace: instance.Namespace,
+			},
+		}
+
+		err = helper.GetClient().Delete(ctx, dnsdata)
+		if err != nil && !k8s_errors.IsNotFound(err) {
+			return ctrl.Result{}, fmt.Errorf("error deleting dnsdata %s: %w", serviceName, err)
 		}
```
stuggi, line 869: cleanup not yet tested

stuggi:
TODO: add tests

controllers/ovndbcluster_controller.go
```diff
@@ -670,135 +670,207 @@ func (r *OVNDBClusterReconciler) reconcileServices(
...
-		// DNSData info is called every reconcile loop to ensure that even if a pod gets
-		// restarted and it's IP has changed, the DNSData CR will have the correct info.
-		// If nothing changed this won't modify the current dnsmasq pod.
-		err = ovndbcluster.DNSData(
+
+		// Create the service
+		svcDef := ovndbcluster.Service(serviceName, instance, serviceLabels, serviceLabels)
+		// make sure that connections from a particular client are passed to the same Pod each time
+		svcDef.Spec.SessionAffinity = corev1.ServiceAffinityClientIP
```
stuggi, line 818: not sure if this is really required

stuggi:
closing and will open a new one

PR #374, opened 2024-11-06
Allow exposing ovn DB using a k8s service

stuggi:
Using multus to attach the ovn db pods to the networks that the edpm nodes can access it won't work for BGP. A loadbalancer service, like metallb, could be used in this case.

This introduces service override to the ovndbcluster, like we have for other services which use a LB service to expose them to the outside. Adding the AnnotationHostnameKey annotation the service hostname gets automatic registered in the DNS instance.

In general this could replace all the current implementation of attaching the ovn db pods using multus, creating the headless service and creating the dnsdata to register it in dns. For update/backward compatability this is kept as if if no service override is provided.

To expose the ovndbcluster-sb via a metallb loadbalancer would look like:

```
  ovn:
    enabled: true
    template:
      ...
      ovnDBCluster:
        ovndbcluster-nb:
	  ...
        ovndbcluster-sb:
          dbType: SB
	  ...
          override:
            service:
              metadata:
                annotations:
                  metallb.universe.tf/address-pool: internalapi
                  metallb.universe.tf/allow-shared-ip: internalapi
                  metallb.universe.tf/loadBalancerIPs: 172.17.0.80
              spec:
                type: LoadBalancer
      ovnNorthd:
        ...
```

Jira: [OSPRH-651](https://issues.redhat.com//browse/OSPRH-651)
related: https://issues.redhat.com/browse/OSPRH-8680
related: https://issues.redhat.com/browse/OSPRH-5504


booxter:
@averdagu since you worked on DNS records for the service, would you mind reviewing this? Thanks.

averdagu:
So far changes looks good to me.
It was a known issue that the current DNSData approach was not working for BGP environments, Thanks for the patch.
Will try to test it locally today afternoon/tomorrow.

stuggi:
just rebased

averdagu:
I tested locally and on a normal environment (non-bgp) seems good, moving from/to lb to normal dnsData works fine. 
lgtm from my part

PR #380, opened 2024-11-19
[ovsdb-server] Gracefully handle empty db file on startup

karelyatin:
In some situations, e.g. when ovsdb-server is killed during db file initialization, the file may end up empty. On next restart, the service fails to start because the file format is invalid.

Ideally, ovs start scripts would handle this situation gracefully, as tracked in https://issues.redhat.com/browse/FDP-689.

But in the meantime, we can also handle the situation on operator's side, by removing the empty file if present. Also added trap signals to the initialization script so it waits for db file to be initialized before getting terminated.

Resolves: OSPRH-10688


karelyatin:
recheck

PR #404, opened 2025-02-12
Consume Topology CR by reference

fmount:
This patch provides more granular control over `Pod` placement and scheduling through the `Topology` CR integration introduced in `infra-operator`.
In particular it provides:

- a new `API` parameter (`TopologyRef`) defined for each `Component` that allows to reference an existing `Topology CR` in the same namespace

- the operator logic that retrieves and processes the referenced `Topology` CR through the functions provided by `infra-operator` `API`

- enhanced `StatefulSet` and `Deployment` configuration that incorporates the processed `Topology`

- a set of `envTest` to test the lifecycle (add/update/override/remove) of the resulting `StatefulSets` when a `Topology` is referenced

Note that `webhooks` are in place to prevent referencing a `Topology` from a different namespace (which is not a supported scenario).

Jira: https://issues.redhat.com/browse/OSPRH-14015


pkg/ovncontroller/daemonset.go
```diff
@@ -150,6 +152,19 @@ func CreateOVNDaemonSet(
 		daemonset.Spec.Template.Spec.NodeSelector = *instance.Spec.NodeSelector
 	}
 
+	if topology != nil {
```
fmount, line 155: @olliewalsh The original `DaemonSet` code does not set anything except `NodeSelector`. It makes sense in this case. and I'm wondering if we prefer to not have topology in this case. In theory it doesn't hurt and `TopologyConstraints` gets ignored, but `NodeAffinity` might be processed [1] (and it's technically part of the `Topology` interface).
I could simplify the code further and remove `TopologySpreadConstraints` because it does not make sense here, but keep affinity because of [1].
I'm going to do this update, but let me know if you have other points we should consider.
 
[1] https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/daemon/util/daemonset_util.go#L226

fmount:
@olliewalsh @stuggi this is ready to go as well

stuggi (approved):
/lgtm

fmount:
```
Unable to connect to the server: dial tcp: lookup api.oko-05.oooci.ccitredhat.com on 172.30.0.10:53: read udp 10.129.68.55:44095->172.30.0.10:53: i/o timeout
````
re-kicking the job

PR #423, opened 2025-03-26
[ovn-controller] Change startup mechanism of ovs pods

averdagu:
This commit aims to modify the starting scripts of the ovn-controller-ovs daemonset.

This is done to allow modifying the RollingUpdate Strategy to not allow any Unavailable pod during update, what this will cause is that during an update to the pod (delete old one and create new one) instead of first deleting the first one and then deleting the next one, it will create the new pod while the old one is running. Due to how the startup scritps works currently this is not allowed.

The reason behind this is to try to lower the downtime observed if the environment is using centralized floating ip during an update.

With this commit the ovn-controller-ovs will share the PID namespace with the host, in order to allow signaling between old/new pod.

Another change is adding an STATE that the containers (ovsdb-server, ovs-vswitchd and ovsdb-server-init) will handle internally.

The differents states are:
 - NULL (No file): will happen the fist time ds is created on the oc worker.
 - INIT: First time init-ovsdb-server is executed on the oc worker.
 - OVSDB_SERVER: once ovsdb-server pod has run the startup script
 - RUNNING: Once ovsdb-server is up  and ovs-vswitchd has run the startup script.
 - UPDATE: Once a new pod is created and ovsdb-server-init has run.
 - RESTART_VSWITCH: After ovsdb-server-init has finished, new ovsdb-server pod has stopped the old ovs-vswitchd process.
 - RESTART_DBSERVER: After old ovs-vswitchd has been restarted the old ovsdb-server is also stop.

The normal flow of states is the following:

NULL -> INIT -> OVSDB_SERVER -> RUNNING

Scale down: If the oc worker is deleted the DS and all the pods and mount points will be deleted, in case of node being up again it should start from NULL

Update: RUNNING -> (Change on CR) -> UPDATE -> RESTART_VSWITCHD ->
RESTART_DBSERVER -> OVSDB_SERVER -> RUNNING

Related: [OSPRH-11636](https://issues.redhat.com//browse/OSPRH-11636)
Jira: [OSPRH-10821](https://issues.redhat.com//browse/OSPRH-10821)
Depends-on: [lib-common#611](https://github.com/openstack-k8s-operators/lib-common/pull/611)

### DIFF
--- a/templates/ovncontroller/bin/init-ovsdb-server.sh
+++ b/templates/ovncontroller/bin/init-ovsdb-server.sh
@@ -18,15 +18,71 @@ set -ex
 source $(dirname $0)/functions
 trap wait_for_db_creation EXIT
 
+function init_ovsdb_server {
+    # Initialize or upgrade database if needed
+    CTL_ARGS="--system-id=random --no-ovs-vswitchd"
+    /usr/share/openvswitch/scripts/ovs-ctl start $CTL_ARGS
+    /usr/share/openvswitch/scripts/ovs-ctl stop $CTL_ARGS
+
+    if [ ! -f /var/lib/openvswitch/already_executed ]; then
+        # If file was not present, set status INIT
+        echo "INIT" > /var/lib/openvswitch/already_executed
+    fi
+
+    wait_for_db_creation
+    trap - EXIT
+}
+
 # If db file is empty, remove it; otherwise service won't start.
 # See https://issues.redhat.com/browse/FDP-689 for more details.
 if ! [ -s ${DB_FILE} ]; then
     rm -f ${DB_FILE}
 fi
-# Initialize or upgrade database if needed
-CTL_ARGS="--system-id=random --no-ovs-vswitchd"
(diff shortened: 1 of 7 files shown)


templates/ovncontroller/bin/init-ovsdb-server.sh
```diff
@@ -18,15 +18,67 @@ set -ex
...
+        # If file was not present, set status INIT
+        echo "INIT" > /var/lib/openvswitch/already_executed
+    fi
+
+    wait_for_db_creation
+    trap - EXIT
+}
+
+TLSOptions="--certificate=/etc/pki/tls/certs/ovndb.crt --private-key=/etc/pki/tls/private/ovndb.key --ca-cert=/etc/pki/tls/certs/ovndbca.crt"
+DBOptions="--db ssl:ovsdbserver-nb.openstack.svc.cluster.local:6641"
```
averdagu, line 37: Only supporting ssl, need to add mechanism to support also non tls connections

PR #428, opened 2025-04-08
Use deployment.IsReady() to validate status

stuggi:
Changes to use the common IsReady() func to validate a deployment is fully up as requested and e.g. no update rollout in progress. During a minor update this has seen to already report/mark the condition ready, even the deployment is still in progress, or the replacement pod failed.

Jira: [OSPRH-14472](https://issues.redhat.com//browse/OSPRH-14472)

Depends-On: https://github.com/openstack-k8s-operators/lib-common/pull/616

### DIFF
--- a/api/go.mod
+++ b/api/go.mod
@@ -7,7 +7,7 @@ require (
 	github.com/google/uuid v1.6.0
 	github.com/onsi/gomega v1.34.1
 	github.com/openstack-k8s-operators/infra-operator/apis v0.6.1-0.20250403063905-eb287d52f38d
-	github.com/openstack-k8s-operators/lib-common/modules/common v0.6.1-0.20250402133843-5a4c5f4fb4f1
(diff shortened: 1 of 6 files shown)


abays (approved):
/lgtm

slawqo (approved):
This change looks good for me but should we also use Statefulset.IsReady() for the ovndbclusters ?

stuggi:
> This change looks good for me but should we also use Statefulset.IsReady() for the ovndbclusters ?

probably yes, I did not do that as part of this PR as there is some special handling in the condition at https://github.com/openstack-k8s-operators/ovn-operator/blob/main/controllers/ovndbcluster_controller.go#L665-L701 , which I am not familiar with if we could just move to 

```
	if statefulset.IsReady(deploy) && len(svcList.Items) > 0 {
```
maybe someone from the team could look into it as a follow up and confirm that this is ok.

PR #430, opened 2025-04-11
Use statefulset.IsReady() to validate status

stuggi:
Changes to use the common IsReady() func to validate a deployment is fully up as requested and e.g. no update rollout in progress. During a minor update this has seen to already report/mark the condition ready, even the deployment is still in progress, or the replacement pod failed.

Jira: OSPRH-14472

Depends-On: https://github.com/openstack-k8s-operators/lib-common/pull/616

### DIFF
--- a/controllers/ovndbcluster_controller.go
+++ b/controllers/ovndbcluster_controller.go
@@ -603,7 +603,10 @@ func (r *OVNDBClusterReconciler) reconcileNormal(ctx context.Context, instance *
 		return ctrlResult, nil
 	}
 
-	instance.Status.ReadyCount = sfset.GetStatefulSet().Status.ReadyReplicas
+	deploy := sfset.GetStatefulSet()
+	if deploy.Generation == deploy.Status.ObservedGeneration {
+		instance.Status.ReadyCount = deploy.Status.ReadyReplicas
+	}
 
 	// verify if network attachment matches expectations
 	networkReady, networkAttachmentStatus, err := nad.VerifyNetworkStatusFromAnnotation(ctx, helper, networkAttachments, serviceLabels, instance.Status.ReadyCount)
@@ -662,7 +665,7 @@ func (r *OVNDBClusterReconciler) reconcileNormal(ctx context.Context, instance *
 		return ctrl.Result{}, err
 	}
 
-	if instance.Status.ReadyCount > 0 && len(svcList.Items) > 0 {
+	if statefulset.IsReady(deploy) && len(svcList.Items) > 0 {
 		instance.Status.Conditions.MarkTrue(condition.DeploymentReadyCondition, condition.DeploymentReadyMessage)
 		instance.Status.Conditions.MarkTrue(condition.ExposeServiceReadyCondition, condition.ExposeServiceReadyMessage)
 		internalDbAddress := []string{}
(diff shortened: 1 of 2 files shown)


stuggi:
@slawqo @karelyatin can you please review if this is appropriate for the ovndbcluster statefulset. I did not fully get the current envtest and the update I did is good.

tests/functional/ovndbcluster_controller_test.go
```diff
@@ -321,12 +321,13 @@ var _ = Describe("OVNDBCluster controller", func() {
 				g.Expect(k8sClient.Update(ctx, c)).Should(Succeed())
 			}).Should(Succeed())
 
-			//Check that error occurs
-			Eventually(func(g Gomega) {
-				conditions := GetOVNDBCluster(dbs[0]).Status.Conditions
-				cond := conditions.Get(condition.ExposeServiceReadyCondition)
-				g.Expect(cond.Status).To(Equal(corev1.ConditionFalse))
-			}).Should(Succeed())
```
stuggi, line 329: this error part is not what I got.

controllers/ovndbcluster_controller.go
```diff
@@ -662,7 +665,7 @@ func (r *OVNDBClusterReconciler) reconcileNormal(ctx context.Context, instance *
 		return ctrl.Result{}, err
 	}
 
-	if instance.Status.ReadyCount > 0 && len(svcList.Items) > 0 {
+	if statefulset.IsReady(deploy) && len(svcList.Items) > 0 {
```
slawqo, line 668: I may be wrong here but I don't really know why we need this additional condition there really. Maybe @karelyatin will know something more about it

slawqo (approved):
lgtm for me

controllers/ovndbcluster_controller.go
```diff
@@ -603,7 +603,10 @@ func (r *OVNDBClusterReconciler) reconcileNormal(ctx context.Context, instance *
 		return ctrlResult, nil
 	}
 
-	instance.Status.ReadyCount = sfset.GetStatefulSet().Status.ReadyReplicas
+	deploy := sfset.GetStatefulSet()
```
karelyatin, line 606: hmm we shoudn't use "deploy" vars for statefulsets :) , we need to cleanup some conditions too for this.

tests/functional/ovndbcluster_controller_test.go
karelyatin: that error was added on purpose some months back by @averdagu , let's check that first

karelyatin (changes requested):

controllers/ovndbcluster_controller.go
karelyatin: We need to ensure statefulset is ready before setting Ready Condition

stuggi: thats what `statefulset.IsReady(deploy)` is checking [1]. it is not just checking that we have >0 replica, it is checking:
```
// IsReady - validates when deployment is ready deployed to whats being requested
// - the requested replicas in the spec matches the ReadyReplicas of the status
// - both when the Generatation of the object matches the ObservedGeneration of the Status
```

[1] https://github.com/openstack-k8s-operators/lib-common/blob/main/modules/common/statefulset/statefulset.go#L142-L145

controllers/ovndbcluster_controller.go
stuggi: I can do that, I kept it as `deploy` since a statefulset is basically also a deployment. let me change it

tests/functional/ovndbcluster_controller_test.go
stuggi: done

controllers/ovndbcluster_controller.go
stuggi: done

stuggi:
rebased

slawqo (approved):

PR #447, opened 2025-06-18
Remove EnableChassisAsGateway option

elvgarrui:
After PR#1070 we don't create ovn-controller if nicMappings is empty. Therefore, there is no need for us to control the EnableChassisAsGateway parameter anymore, since functionalities that require external connectivity do require nicMappings anyway.

Resolves: OSPRH-7463


api/v1beta1/ovncontroller_types.go
```diff
@@ -174,10 +174,6 @@ type OVSExternalIDs struct {
 	// +kubebuilder:validation:Optional
 	// +kubebuilder:default={}
 	OvnAvailabilityZones []string `json:"availability-zones,omitempty"`
-
-	// +kubebuilder:validation:Optional
-	// +kubebuilder:default=true
-	EnableChassisAsGateway *bool `json:"enable-chassis-as-gateway"`
```
karelyatin, line 180: crd-checker looks failing so need to see what's proper way for removing fields.

elvgarrui:
If I understood correctly from https://book.kubebuilder.io/multiversion-tutorial/api-changes, we might need to create a new API version in order to be able to remove the parameter completely. Before that, it is very encouraged to deprecate it in this version, hence the change of content on this commit. Should I proceed by uploading this first and then a second PR where a new API version is created without this field? @karelyatin Thanks for your inputs!

elvgarrui:
> > If I understood correctly from https://book.kubebuilder.io/multiversion-tutorial/api-changes, we might need to create a new API version in order to be able to remove the parameter completely. Before that, it is very encouraged to deprecate it in this version, hence the change of content on this commit. Should I proceed by uploading this first and then a second PR where a new API version is created without this field? @karelyatin Thanks for your inputs!
> 
> @elvgarrui from where you got kubebuilder:validation:Deprecated ? i didn't found anything about it. Not sure if we need to do change versions just for this have to dig. But what i think should be fine for now is original commit just marking the field deprecation via description and not honor spec.externalIds.enableChassisAsGateway and default to true

You are right, I got confused with // +kubebuilder:deprecatedversion in https://book.kubebuilder.io/reference/markers/crd and thought it would be possible to apply it to independent values. But it isn't.

PR #451, opened 2025-06-25
Deprecate EnableChassisAsGateway option

elvgarrui:
After ovn-operator/PR#1070 we don't create ovn-controller CR if nicMappings is empty. Therefore, there is no need for us to expose the EnableChassisAsGateway parameter anymore, since functionalities that require external connectivity do require nicMappings anyway. Since we cannot remove the definition from the API without changing the version, we should at least deprecate it.

Related: OSPRH-7463


api/v1beta1/ovncontroller_types.go
```diff
@@ -175,6 +175,7 @@ type OVSExternalIDs struct {
 	// +kubebuilder:default={}
 	OvnAvailabilityZones []string `json:"availability-zones,omitempty"`
 
+	// This field is deprecated and might be removed when a new api version is released
```
karelyatin, line 178: What about changing this to?
        // DEPRECATED: To be removed in the next API version
        // Any value set to this field is ignored

karelyatin (approved):

PR #456, opened 2025-07-04
Change liveness and readiness probes for the ovn-northd POD

slawqo:
Before this patch liveness and readiness probes for the ovn-northd POD were just checking PID of the ovn-northd process.

This patch changes it to use custom script which checks ovn-northd status with `ovs-appctl status` command.
According to the OVN documentation [1] this command:

  Prints this server’s status. Status will be "active" if ovn-northd has
  acquired OVSDB lock on SB DB, "standby" if it has not or "paused" if
  this instance is paused.

As `ovs-appctl status` command returns `0` as exit code even if status of the ovn-northd is e.g. "standby", this patch adds custom script which is used by the liveness and readiness probes and which returns `1` in case when status is anything else then `active`.

[1] https://man7.org/linux/man-pages/man8/ovn-northd.8.html

Closes: #OSPRH-8042


averdagu:
Seems good, let's wait for all checks to be green.
Question, IIRC when the liveness/probeness of the ovn-controller [0] was added, also some kuttl tests were created, should we need a kuttl test in this case?
IMHO it's a simple script and shouldn't, just popping the question just in case

[0] https://github.com/openstack-k8s-operators/ovn-operator/pull/360/

templates/ovnnorthd/bin/status_check.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+NORTHD_PID=$(/usr/bin/pidof ovn-northd)
+UNIXCTL_SOCKET="${OVN_RUNDIR}/ovn-northd.${NORTHD_PID}.ctl"
+
+status=$(/usr/bin/ovs-appctl -t $UNIXCTL_SOCKET status)
+if [[ $status != *"active"* ]]; then
+    exit 1
```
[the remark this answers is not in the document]
slawqo: You are totally right, I was testing it only on the env with one northd POD. I now changed it to use `ovn-appctl {nb,sb}-connection-status` which should be better. Please let me know what do you think about it.

templates/ovnnorthd/bin/status_check.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+
+# Script returns:
+# 0 - if both NB and SB connections are OK
+# 1 - if there is no NB connection
+# 2 - if there is no SB connection
+# 3 - if there is no both NB and SB connection
```
karelyatin, line 7: if there are no NB and SB connections

templates/ovnnorthd/bin/status_check.sh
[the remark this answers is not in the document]
karelyatin: Thanks for checking, can you also extend the kuttl tests to cover this scenario, i.e deploy 2 replicas of northd.

controllers/ovnnorthd_controller.go
```diff
@@ -445,6 +447,24 @@ func (r *OVNNorthdReconciler) reconcileNormal(ctx context.Context, instance *ovn
 	// all cert input checks out so report InputReady
 	instance.Status.Conditions.MarkTrue(condition.TLSInputReadyCondition, condition.InputReadyMessage)
 
+	//
+	// create Configmap required for the Northd service
+	// - %-config configmap holding scripts required for the Northd service
```
karelyatin, line 452: -scripts configmap

controllers/ovnnorthd_controller.go
```diff
@@ -445,6 +447,24 @@ func (r *OVNNorthdReconciler) reconcileNormal(ctx context.Context, instance *ovn
...
+	if err != nil {
+		instance.Status.Conditions.Set(condition.FalseCondition(
+			condition.ServiceConfigReadyCondition,
+			condition.ErrorReason,
+			condition.SeverityWarning,
+			condition.ServiceConfigReadyErrorMessage,
+			err.Error()))
+		return ctrl.Result{}, err
+	}
+	// Create ConfigMaps and Secrets - end
```
karelyatin, line 464: /and Secrets/d

pkg/ovnnorthd/volumes.go
```diff
@@ -0,0 +1,35 @@
...
+						Name: name + "-scripts",
+					},
+				},
+			},
+		},
+	}
+
+}
+
+// GetNorthdVolumeMounts - OVN DBCluster VolumeMounts
```
karelyatin, line 25: s/DBCluster/Northd/

PR #464, opened 2025-07-29
bump rabbitmq-cluster-operator

stuggi:


abays (approved):
/lgtm

karelyatin:
actually we don't need cluster-operator it in ovn so can just cleanup

stuggi:
> actually we don't need cluster-operator it in ovn so can just cleanup

we need it because how the indirect imports we get due to our operator dependencies.

PR #472, opened 2025-08-28
Enable exporter for northd

karelyatin:
When spec.metricsEnabled=true(default) a container exporting metrics is added to the Northd pod. Services per pod are created which will be used by telemetry prometheus instance to pull these metrics.

TODO: add kuttl tests

Assisted-by: Claude


karelyatin:
recheck crc

karelyatin:
in favor of https://github.com/openstack-k8s-operators/ovn-operator/pull/475

PR #490, opened 2025-09-26
Support bond configuration in NicMappings

elvgarrui:
This patch introduces the possibility of defining a new OVNController API parameter called BondConfiguration which is used together with NicMappings to allow the existence bond type interfaces.

The configuration would be structured as follows:

spec:
...
  bondConfiguration:
    bond0:
      links:
      - eth10.10
      - eth10.20
      mtu: 1500
      mode: active-backup
  nicMappings:
    datacentre: bond0
...

In this case, the datacentre nic would be the primary bond interface and links eth10.10 and eth10.20 the members of such configuration. Mode and MTU can be optionally changed too, with defaults set as the one seen in the example above.


api/v1beta1/ovncontroller_types.go
```diff
@@ -75,6 +75,11 @@ type OVNControllerSpecCore struct {
 	// +optional
 	NicMappings map[string]string `json:"nicMappings,omitempty"`
 
+	// +kubebuilder:validation:Optional
+	// +optional
+	// Map of bond names to their configuration (links and mode)
```
karelyatin, line 80: can remove links and mode from the description, else we have to add mtu or any other future bond option here.

pkg/ovncontroller/network.go
```diff
@@ -15,82 +15,184 @@ package ovncontroller
...
+	"golang.org/x/exp/maps"
 	k8s_errors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	ovnv1 "github.com/openstack-k8s-operators/ovn-operator/api/v1beta1"
 )
 
-// CreateOrUpdateAdditionalNetworks - create or update network attachment definitions based on the provided mappings
-func CreateOrUpdateAdditionalNetworks(
```
karelyatin, line 29: Let's also have kuttl tests to validate this new config, can extend test ovn_nicmappings itself

controllers/ovncontroller_controller.go
```diff
@@ -495,7 +494,17 @@ func (r *OVNControllerReconciler) reconcileNormal(ctx context.Context, instance
...
+	// network to attach to
+	networkAttachments := []string{}
+	networkAttachmentsNoPhysNet := []string{}
+	if instance.Spec.NetworkAttachment != "" {
+		networkAttachments = append(networkAttachments, instance.Spec.NetworkAttachment)
+		networkAttachmentsNoPhysNet = append(networkAttachmentsNoPhysNet, instance.Spec.NetworkAttachment)
+	}
+
+	networkAttachmentsToAdd, err := ovncontroller.CreateOrUpdateAdditionalNetworks(ctx, helper, instance, ovsServiceLabels, networkAttachments)
+	networkAttachments = append(networkAttachments, networkAttachmentsToAdd...)
```
karelyatin, line 506: This changing the order of interfaces so during update to this version of operator ovs pods will be restarted, which we should avoid.

karelyatin: Yes order is changed, for instance networkAttachment=tenant and nicMappings=datacentre:enp8s0
Before this patch nic order was datacentre, tenant
And now with this patch order is tenant, datacentre
So if you can just reorder(networkAttachmentsToAdd + networkAttachments) the append in L506 it should be enough to persist original order.

elvgarrui: Done!

karelyatin:
> I don't fully understand why this is failing on the gate. It passes locally

It's because of hardcoded namespace "ovn-kuttl" in 06-assert.yaml

api/v1beta1/ovncontroller_types.go
```diff
@@ -75,6 +75,11 @@ type OVNControllerSpecCore struct {
 	// +optional
 	NicMappings map[string]string `json:"nicMappings,omitempty"`
 
+	// +kubebuilder:validation:Optional
+	// +optional
+	// Map of bond names to their configuration
+	BondConfiguration map[string]Bond `json:"bondConfiguration,omitempty"`
```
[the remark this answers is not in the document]
stuggi: perfect, missed that bit. if we have that possibility to use a pre-created NAD, just wondering if we need this enhancement to the CRD, or just document to create it as needed.

pkg/ovncontroller/network.go
```diff
@@ -15,82 +15,185 @@ package ovncontroller
...
+		// If the name of the interface is configured as bond, add bond NAD instead of the default nic NAD.
+		interfaceName := instance.Spec.NicMappings[physNet]
+		var nadConfig string
+		if slices.Contains(bondKeys, interfaceName) {
+			linkNames, err := createLinkNADs(ctx, h, instance, labels, instance.Spec.BondConfiguration[interfaceName].Links)
+			if err != nil {
+				return nil, err
 			}
+			finalNetworkAttachments = append(finalNetworkAttachments, linkNames...)
+			nadConfig = defineBondNADConfig(physNet, instance.Spec.BondConfiguration[interfaceName].Mode, instance.Spec.BondConfiguration[interfaceName].Mtu, linkNames)
```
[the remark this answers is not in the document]
elvgarrui: I still think leaving linkNames as a separate list is better though because the name of the NADs might not be the same as the name of the plain link, so now it is  `nadConfig = defineBondNADConfig(physNet, instance.Spec.BondConfiguration[interfaceName], linkNames)`

hack/crd-schema-checker.sh
```diff
@@ -28,5 +28,7 @@ for crd in config/crd/bases/*.yaml; do
     $CHECKER check-manifests \
         $CHECKER_ARGS \
         --existing-crd-filename="$TMP_DIR/$crd" \
-        --new-crd-filename="$crd"
+        --new-crd-filename="$crd" \
+        --disabled-validators NoMaps \
+        --disabled-validators NoNewRequiredFields
```
stuggi, line 33: we should not disable it. if there is a new required field and its ok, we should instead override the failing CI job. otherwise it would need a next PR to enable it again.

elvgarrui: Ok np! I reverted this bit and validated the links from the controller itself.

PR #498, opened 2025-10-06
Change default namespace for kuttl tests

elvgarrui:
Previously we were using the production namespace, "openstack". This is not convenient since it can overlap with pods that are from a development or production environment and it would be easier to have the same name as the one configured in install_yamls[0].

[0] https://github.com/openstack-k8s-operators/install_yamls/blob/main/Makefile#L236


karelyatin (approved):
/lgtm

PR #541, opened 2026-03-20
Use RBAC while connecting ovn-controllers to SB database

slawqo:
This patch configures RBAC to access OVN SB databases so that ovn-controllers now have limited access to this DB and will only be able to modify its own data.

On the other hand Northd requires "full access" to the SB DB, and to achieve that there is another DB listener created on port 16642 for to be used by northd.

More info about OVN RBAC can be found in its documentation at [1].

[1] https://docs.ovn.org/en/latest/tutorials/ovn-rbac.html

Depends-On: https://github.com/openstack-k8s-operators/install_yamls/pull/1145

Related: #[OSPRH-1921](https://redhat.atlassian.net/browse/OSPRH-1921)
Closes: #[OSPRH-1922](https://redhat.atlassian.net/browse/OSPRH-1922)

### DIFF
--- a/api/bases/ovn.openstack.org_ovndbclusters.yaml
+++ b/api/bases/ovn.openstack.org_ovndbclusters.yaml
@@ -430,6 +430,12 @@ spec:
                 description: InternalDBAddress - DB IP address used by other Pods
                   in the cluster
                 type: string
+              internalDbAddressRbacFullAccess:
+                description: |-
+                  InternalDBAddressRbacFullAccess - DB IP address used by other Pods which
+                  requires full access to the SB db, like e.g. Northd. This is used only
+                  when OVN RBAC for ovn-controllers is used (TLS enabled)
+                type: string
               lastAppliedTopology:
                 description: LastAppliedTopology - the last applied Topology
                 properties:
--- a/api/test/helpers/crd.go
+++ b/api/test/helpers/crd.go
@@ -130,6 +130,7 @@ func (th *TestHelper) SimulateOVNNorthdReady(name types.NamespacedName) {
 	gomega.Eventually(func(g gomega.Gomega) {
 		service := th.GetOVNNorthd(name)
 		service.Status.ObservedGeneration = service.Generation
+		service.Status.ReadyCount = 1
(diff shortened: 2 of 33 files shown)


stuggi:
* do the ovn-controllers on the edpm nodes also need a rbac cert? how would they get it?
* right now all CA certs/service certs are managed by the openstack-op, either ctlplane cotnroller, or dataplanecontroller,  and services just consume the generated. not sure if we could keep that? or if it could work. there also certs for edpm nodes get created/requested.
* how would backup/restore work for this. just thinking about it as I work on b/r for the ctlplane (except ovn). we have to think about this.

slawqo:
> * do the ovn-controllers on the edpm nodes also need a rbac cert? how would they get it?

Yes, they need it too. I proposed to mount secret with that SB Cert to the ansibleee pod and then generate SSL cert for each node in that Ansibleee POD and copy it to the edpm node. PRs for that part are https://github.com/openstack-k8s-operators/openstack-operator/pull/1895 and https://github.com/openstack-k8s-operators/edpm-ansible/pull/1158

> 
>     * right now all CA certs/service certs are managed by the openstack-op, either ctlplane cotnroller, or dataplanecontroller,  and services just consume the generated. not sure if we could keep that? or if it could work. there also certs for edpm nodes get created/requested.

Are you saying that I should generate all certs by openstack-operator or that I should do that only for the SB Cert which is later used to sign certificates for OVN controller (PODs and services running on edpm nodes)? The problem with the latter is that certificates used by ovn-controller have to have CN field which matches "system-id" from that node and have to be signed with cert generated on the SB DB POD. So how it could all be handled by openstack-operator? Should we generate uuids for ovn-controller PODs there too?

> 
>     * how would backup/restore work for this. just thinking about it as I work on b/r for the ctlplane (except ovn). we have to think about this.

I didn't test backup/restore really but I'm not sure what problem we could have with it. Isn't backup/restore already doing backup of certificates on the nodes?

slawqo:
Replaced by new approach proposed in https://github.com/openstack-k8s-operators/ovn-operator/pull/565

PR #549, opened 2026-04-02
[ovndbcluster] Allow config changes without pod restarts

karelyatin:
Currently for any config change pod get's restart which we don't need for configs like:-
- ElectionTimer
- InactivityProbe
- LogLevel

Deprecated probeIntervalToActive config option as
that is not relevant only for active/standby db clusters which we are not using here.

Also using db directories(as persistent) for OVN_RUNDIR, so we can use them in job pods to do the config changes.

Resolves: #[OSPRH-27985](https://redhat.atlassian.net/browse/OSPRH-27985)
Assisted-By: Claude

### DIFF
--- a/internal/controller/ovndbcluster_controller.go
+++ b/internal/controller/ovndbcluster_controller.go
@@ -46,6 +46,7 @@ import (
 	"github.com/openstack-k8s-operators/lib-common/modules/common/configmap"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/env"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/helper"
+	"github.com/openstack-k8s-operators/lib-common/modules/common/job"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/labels"
 	nad "github.com/openstack-k8s-operators/lib-common/modules/common/networkattachment"
 	common_rbac "github.com/openstack-k8s-operators/lib-common/modules/common/rbac"
@@ -562,6 +563,7 @@ func (r *OVNDBClusterReconciler) reconcileNormal(ctx context.Context, instance *
 	// create Configmap required for dbcluster input
 	// - %-config configmap holding minimal dbcluster config required to get the service up
 	//
+	// Create ConfigMaps with ALL parameters included for setup script (pods will use these)
 	err = r.generateServiceConfigMaps(ctx, helper, instance, &configMapVars, serviceName)
 	if err != nil {
 		instance.Status.Conditions.Set(condition.FalseCondition(
@@ -660,6 +662,21 @@ func (r *OVNDBClusterReconciler) reconcileNormal(ctx context.Context, instance *
 	stateful := sfset.GetStatefulSet()
 	if stateful.Generation == stateful.Status.ObservedGeneration {
 		instance.Status.ReadyCount = stateful.Status.ReadyReplicas
+	}
+
+	// Only run runtime config when all pods are ready and no rolling update in progress
+	if instance.Status.ReadyCount == *instance.Spec.Replicas &&
+		stateful.Status.UpdatedReplicas == *instance.Spec.Replicas {
+		err = r.reconcileRuntimeConfig(ctx, helper, instance, serviceLabels, serviceName)
+		if err != nil {
+			instance.Status.Conditions.Set(condition.FalseCondition(
+				condition.ServiceConfigReadyCondition,
+				condition.ErrorReason,
+				condition.SeverityWarning,
+				condition.ServiceConfigReadyErrorMessage,
+				err.Error()))
+			return ctrl.Result{}, err
+		}
 	}
 
 	// verify if network attachment matches expectations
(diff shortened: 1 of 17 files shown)


karelyatin:
recheck trunkrdo

internal/controller/ovndbcluster_controller.go
```diff
@@ -573,11 +575,32 @@ func (r *OVNDBClusterReconciler) reconcileNormal(ctx context.Context, instance *
 		return ctrl.Result{}, err
 	}
 
+	// Create SEPARATE environment variables for hash calculation WITHOUT configurable parameters
+	// This ensures election timer, log level, and inactivity probe changes don't trigger pod restarts
```
averdagu: maybe a better wording would be replace "ensures" with "includes"?

internal/controller/ovndbcluster_controller.go
```diff
@@ -1112,6 +1151,135 @@ func (r *OVNDBClusterReconciler) deleteExternalConfigMaps(
...
+) error {
+	Log := r.GetLogger(ctx)
+
+	// Calculate combined hash for all runtime parameters
+	runtimeConfigHash, err := util.ObjectHash(map[string]interface{}{
+		"ElectionTimer":   instance.Spec.ElectionTimer,
+		"InactivityProbe": instance.Spec.InactivityProbe,
+		"LogLevel":        instance.Spec.LogLevel,
+		"DBType":          instance.Spec.DBType,
+		"TLS":             instance.Spec.TLS.Enabled(),
```
averdagu: TLS could be removed from here, as it won't affect for the Runtime parameters, and it could happen that if you only modify the TLS it would trigger a reconcileRuntimeConfig but nothing will change.

internal/controller/ovndbcluster_controller.go
```diff
@@ -1112,6 +1151,135 @@ func (r *OVNDBClusterReconciler) deleteExternalConfigMaps(
...
+	// Check if any runtime config changed
+	runtimeConfigHashKey := "OvnDBClusterRuntimeConfigHash"
+	currentHash := instance.Status.Hash[runtimeConfigHashKey]
+
+	if runtimeConfigHash == currentHash {
+		// No changes detected
+		return nil
+	}
+
+	Log.Info("Runtime configuration changed",
```
averdagu, line 1163: NIT: for better debugging, should we also add a log stating the new values?

averdagu: I think it could be useful if customer modifies election timer or inactivity probe, and sets an incorrect value

internal/ovndbcluster/runtimeconfigjob.go
```diff
@@ -0,0 +1,191 @@
...
+	ctx context.Context,
+	k8sClient client.Client,
+	instance *ovnv1.OVNDBCluster,
+	labels map[string]string,
+	serviceName string,
+	changedConfigs map[string]interface{}, // Map of changed configuration parameters
+) ([]*batchv1.Job, error) {
+
+	var jobs []*batchv1.Job
+	// NOTE: set TTLSecondsAfterFinished=0 will clean done
```
averdagu, line 41: This is something that we already use in ovn-operator, but I think it's better to not include the jobTTLAfterFinished, and whenever we need to create a new job check if we already have a job's CR and delete it.
The reason behind that is  that you can't see the last job executed since the CR will be deleted due to that.
Just adding this as a possible discussion. Since it's what we already use I'm fine with leaving it like that

averdagu:
So far code looks good to me, I'll test this tomorrow

internal/controller/ovndbcluster_controller.go
karelyatin: But that doesn't reflect full statement as emphases on ensuring pod don't restart on these runtime config changes.

internal/controller/ovndbcluster_controller.go
karelyatin: Done, also removed DBType

internal/controller/ovndbcluster_controller.go
karelyatin: Done

internal/ovndbcluster/runtimeconfigjob.go
karelyatin: Ok then we can leave it for later and if needed sync at other places

internal/controller/ovndbcluster_controller.go
averdagu: True, I guess I missread the comment, sorry

internal/ovndbcluster/runtimeconfigjob.go
averdagu: Cool

internal/ovndbcluster/runtimeconfigjob.go
```diff
@@ -0,0 +1,191 @@
+/*
```
slawqo, line 1: is name of this file correct? It suggests that it will only configure "election_timer" option. But I think it is used to set all "runtime config options", isn't it?

karelyatin: yes true Thanks, when i started with this patch only targetted election timer and later applied to all configs and missed the rename.

averdagu:
Adding comment here instead of code as I don't know the best way to handle this, but we need to sanitize the input given by the user.
I tested out by adding "logLevel: debug" but ovn-appctl only accepts "dbg". 
We need to check the input either on the CR, or in the script (though doing it in the script may seem all went fine but log level was not changed)

karelyatin:
> Adding comment here instead of code as I don't know the best way to handle this, but we need to sanitize the input given by the user. I tested out by adding "logLevel: debug" but ovn-appctl only accepts "dbg". We need to check the input either on the CR, or in the script (though doing it in the script may seem all went fine but log level was not changed)

yes i noticed that too but since it was not specific to this change i didn't touched api for this. So we can take that as follow up too

averdagu:
> > Adding comment here instead of code as I don't know the best way to handle this, but we need to sanitize the input given by the user. I tested out by adding "logLevel: debug" but ovn-appctl only accepts "dbg". We need to check the input either on the CR, or in the script (though doing it in the script may seem all went fine but log level was not changed)
> 
> yes i noticed that too but since it was not specific to this change i didn't touched api for this. So we can take that as follow up too

For the record, created an issue for this problem that I plan to work it on :)

averdagu:
Looks all good to me

slawqo (approved):

karelyatin:
must gather fail

karelyatin:
gather failure last

PR #559, opened 2026-04-20
Add OVNDBBackup and OVNDBRestore CRDs for managed backup/restore

lmiccini:
Introduces two new Custom Resource Definitions for automated OVN database backup and restore operations:

OVNDBBackup:
- Scheduled backups via CronJob using ovsdb-client backup
- Configurable retention policy
- TLS support for database connections
- Persistent storage for backup files (survives CR deletion)

OVNDBRestore:
- Phase-based state machine: Validating → ScalingDown → Restoring → ScalingUp → Completed
- Annotation-based replica override prevents higher-level operators from interfering during restore
- Force-deletes pods during scale-down (preStop hooks hang when all RAFT members terminate simultaneously)
- Deletes PVCs to prevent stale RAFT membership state
- Copies standalone backup to pod-0 PVC and lets ovn-ctl handle the standalone-to-clustered conversion on startup
- Controlled scale-up: pod-0 first, then remaining replicas
- Post-restore DB verification via exec

Also includes:
- Functional tests for both controllers
- Webhooks with validation
- Documentation


karelyatin:
@lmiccini can you add some more context/Jira here as also noticed some other backup related effort as part of https://github.com/openstack-k8s-operators/ovn-operator/pull/557

lmiccini:
> @lmiccini can you add some more context/Jira here as also noticed some other backup related effort as part of #557

Hey Yatin, yes this is to provide an API around that script, following the mariadb-operator approach. This way we can orchestrate backups and restores.

lmiccini:
recheck

lmiccini:
recheck

otherwiseguy:
I have not yet read the patch, but I'm not sure what the use case is for scheduled ovsdb backups. We already have an active-active cluster, and aside from non-changing initial configuration data the entire database can be generated from the neutron db and ovn-northd. In addition, *any* backup, will be out of date and need to be synced with the neutron db--basically the same process as not having a backup.

PR #562, opened 2026-04-21
Sanitize logLevel CR field

averdagu:
Currently log level can be set to any string value, problem is that it only accepts a set of strings (which are only on the description) but user can set this values to whatever he wants, and could make the operator not being able to spawn all needed pods.

Allowing only to add the correct values

Resolves: [OSPRH-28896](https://redhat.atlassian.net/browse/OSPRH-28896)

### DIFF
--- a/api/v1beta1/ovncontroller_types.go
+++ b/api/v1beta1/ovncontroller_types.go
@@ -95,13 +95,15 @@ type OVNControllerSpecCore struct {
 	NetworkAttachment string `json:"networkAttachment"`
 
 	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Enum=off;emer;err;warn;info;dbg
 	// +kubebuilder:default=info
-	// OVNLogLevel - Set log level off, emer, err, warn, info, or dbg. Default is info.
+	// OVNLogLevel - Set log level off, emer, err, warn, info or dbg. Default is info.
 	OVNLogLevel string `json:"ovnLogLevel,omitempty"`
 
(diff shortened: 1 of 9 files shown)


api/v1beta1/ovndbcluster_types.go
```diff
@@ -84,6 +84,7 @@ type OVNDBClusterSpecCore struct {
 	NodeSelector *map[string]string `json:"nodeSelector,omitempty"`
 
 	// +kubebuilder:validation:Optional
+    // +kubebuilder:validation:Enum=off;emer;err;warn;info;dbg
```
karelyatin, line 87: can you also update below description to the values added in this Enum along with what is default

api/v1beta1/ovncontroller_types.go
```diff
@@ -95,11 +95,13 @@ type OVNControllerSpecCore struct {
 	NetworkAttachment string `json:"networkAttachment"`
 
 	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Enum=off;emer;err;warn;info;dbg
 	// +kubebuilder:default=info
 	// OVNLogLevel - Set log level off, emer, err, warn, info, or dbg. Default is info.
 	OVNLogLevel string `json:"ovnLogLevel,omitempty"`
 
 	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Enum=off;emer;err;warn;info;dbg
```
karelyatin, line 104: Also add what is default in the description

karelyatin (approved):
/lgtm

PR #565, opened 2026-04-29
Add OVN RBAC support with per-node ovn-controller certificates

slawqo:
Enable OVN role-based access control (RBAC) on the Southbound database so that ovn-controller nodes can only modify their own chassis rows.

When the openstack-operator provides an RBAC issuer name (from a dedicated rootca-ovn-rbac CA, see patch [1]), this patch:

* Creates per-node cert-manager Certificate CRs for each ovn-controller pod, with CN set to a deterministic UUID5 system-id derived from the node name (ComputeSystemID). This CN must match the chassis system-id for RBAC to authorize operations.

* Copies the RBAC client cert/key into /etc/openvswitch/ on each node via the config job, and switches ovn-controller to use these dedicated paths instead of the shared OVN DB cert.

* Mounts the RBAC CA certificate into ovsdbserver-sb pods and builds a combined CA bundle (regular CA + RBAC CA) so the SB database can verify ovn-controller client certificates.

* Sets role=ovn-controller on the SB DB connection (port 6642) to enforce RBAC.

* Creates a second SB DB listener on port 16642 with full (unrestricted) access, used by ovn-northd.

* Updates inactivity probe handling in setup.sh and runtime-config.sh to iterate over all connections, since SB now has two listeners.

[1] https://github.com/openstack-k8s-operators/openstack-operator/pull/1906

Depends-On: https://github.com/openstack-k8s-operators/openstack-operator/pull/1906

Related: #[OSPRH-1921](https://redhat.atlassian.net/browse/OSPRH-1921)

Closes: #[OSPRH-1922](https://redhat.atlassian.net/browse/OSPRH-1922)

### DIFF
--- a/internal/controller/ovncontroller_controller.go
+++ b/internal/controller/ovncontroller_controller.go
@@ -37,7 +37,9 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
+	certmgrv1 "github.com/cert-manager/cert-manager/pkg/apis/certmanager/v1"
 	topologyv1 "github.com/openstack-k8s-operators/infra-operator/apis/topology/v1beta1"
+	certmanager "github.com/openstack-k8s-operators/lib-common/modules/certmanager"
 	"github.com/openstack-k8s-operators/lib-common/modules/common"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/condition"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/configmap"
@@ -96,6 +98,9 @@ func (r *OVNControllerReconciler) GetClient() client.Client {
 // +kubebuilder:rbac:groups="security.openshift.io",resourceNames=anyuid;privileged,resources=securitycontextconstraints,verbs=use
 // +kubebuilder:rbac:groups="",resources=pods,verbs=create;delete;get;list;patch;update;watch
 // +kubebuilder:rbac:groups=topology.openstack.org,resources=topologies,verbs=get;list;watch;update
+// cert-manager permissions for per-node RBAC certificates
+// +kubebuilder:rbac:groups=cert-manager.io,resources=certificates,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=cert-manager.io,resources=issuers,verbs=get;list;watch
 
 // Reconcile reconciles the OVNController CR to deploy OVN controller pods
 func (r *OVNControllerReconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ctrl.Result, _err error) {
@@ -266,7 +271,9 @@ func (r *OVNControllerReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Owns(&corev1.ServiceAccount{}).
 		Owns(&rbacv1.Role{}).
 		Owns(&rbacv1.RoleBinding{}).
+		Owns(&certmgrv1.Certificate{}).
 		Watches(&ovnv1.OVNDBCluster{}, handler.EnqueueRequestsFromMapFunc(ovnv1.OVNCRNamespaceMapFunc(crs, mgr.GetClient()))).
+		Watches(&ovnv1.OVNNorthd{}, handler.EnqueueRequestsFromMapFunc(ovnv1.OVNCRNamespaceMapFunc(crs, mgr.GetClient()))).
 		Watches(
 			&corev1.Secret{},
 			handler.EnqueueRequestsFromMapFunc(r.findObjectsForSrc),
@@ -641,6 +648,26 @@ func (r *OVNControllerReconciler) reconcileNormal(ctx context.Context, instance 
 		return ctrl.Result{}, fmt.Errorf("waiting for Topology requirements: %w", err)
 	}
 
+	// When TLS is enabled, RBAC is enforced on the SB database. The RBAC
+	// permission tables are populated by ovn-northd, so ovn-controller must
+	// not start until northd is running; otherwise all SB operations are
+	// denied with "permission error".
+	if instance.Spec.TLS.Enabled() {
+		northd, err := ovnv1.GetOVNNorthd(ctx, helper, instance.Namespace)
+		if err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to look up OVNNorthd: %w", err)
+		}
+		if northd == nil || northd.Status.ReadyCount == 0 {
+			Log.Info("OVNNorthd is not ready yet, waiting before deploying OVNController...")
+			instance.Status.Conditions.Set(condition.FalseCondition(
+				condition.DeploymentReadyCondition,
+				condition.RequestedReason,
+				condition.SeverityInfo,
+				condition.DeploymentReadyRunningMessage))
+			return ctrl.Result{RequeueAfter: time.Duration(5) * time.Second}, nil
+		}
+	}
+
 	// Define a new DaemonSet object for OVNController
 	dset := daemonset.NewDaemonSet(
 		ovncontroller.CreateOVNDaemonSet(instance, inputHash, ovnServiceLabels, topology),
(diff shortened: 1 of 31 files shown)


internal/ovncontroller/daemonset.go
```diff
@@ -59,9 +59,14 @@ func CreateOVNDaemonSet(
...
+		mounts = append(mounts, corev1.VolumeMount{
+			Name:      "etc-ovs",
+			MountPath: "/etc/openvswitch",
+			ReadOnly:  true,
+		})
 		cmd = append(cmd, []string{
-			fmt.Sprintf("--certificate=%s", ovn_common.OVNDbCertPath),
-			fmt.Sprintf("--private-key=%s", ovn_common.OVNDbKeyPath),
+			fmt.Sprintf("--certificate=%s", ovn_common.OVNControllerCertPath),
+			fmt.Sprintf("--private-key=%s", ovn_common.OVNControllerKeyPath),
```
stuggi, line 70: Can we expect RBAC is always enabled? Without RBAC configured (RbacIssuerName empty), no config job copies certs to these paths. This should probably be conditional:                                          

```
  if instance.Spec.RbacIssuerName != "" {
      cmd = append(cmd, fmt.Sprintf("--certificate=%s", ovn_common.OVNControllerCertPath))
      cmd = append(cmd, fmt.Sprintf("--private-key=%s", ovn_common.OVNControllerKeyPath))  
  } else {                                                                                    
      cmd = append(cmd, fmt.Sprintf("--certificate=%s", ovn_common.OVNDbCertPath))            
      cmd = append(cmd, fmt.Sprintf("--private-key=%s", ovn_common.OVNDbKeyPath))             
  } 
```

internal/ovndbcluster/service.go
```diff
@@ -38,6 +38,15 @@ func Service(
 		},
 	}
 
+	// Add full-access (non-RBAC) port for SB when TLS is enabled
+	if instance.Spec.DBType == ovnv1.SBDBType && instance.Spec.TLS.Enabled() {
```
stuggi, line 42: maybe should also be conditional on `RbacCACertSecretName != ""`, so only if rbac is enabled.

templates/ovndbcluster/bin/setup.sh
```diff
@@ -122,15 +129,26 @@ if [[ "$(hostname)" == "{{ .SERVICE_NAME }}-0" ]]; then
     export OVN_${DB_TYPE^^}_DAEMON=$(${CTLCMD} --pidfile --detach)
 
 {{- if .TLS }}
-    ${CTLCMD} set-ssl {{.OVNDB_KEY_PATH}} {{.OVNDB_CERT_PATH}} {{.OVNDB_CACERT_PATH}}
+    ${CTLCMD} set-ssl {{.OVNDB_KEY_PATH}} {{.OVNDB_CERT_PATH}} ${SSL_CA_CERT}
+    if [[ "${DB_TYPE}" == "sb" ]]; then
+        # Use RBAC for the connections of the ovn-controller to SB
+        ${CTLCMD} set-connection role=ovn-controller ${DB_SCHEME}:${DB_PORT}:${DB_ADDR}
```
stuggi, line 130: RBAC role is set on all SB databases when TLS is enabled, even if no RBAC certs exist. Combined with issue mentioned above if RBAC is not enabled (CA set), this means TLS-enabled deployments without RBAC will have the SB reject all ovn-controller connections.
Maybe we should do this?
```
  {{- if index . "OVN_RBAC_CACERT_PATH" }}
      ${CTLCMD} set-connection role=ovn-controller ${DB_SCHEME}:${DB_PORT}:${DB_ADDR}                                                                                                        
      ...                                                                                                                                     
  {{- else }}                                                                                                                                                                                
      ${CTLCMD} set-connection ${DB_SCHEME}:${DB_PORT}:${DB_ADDR}
  {{- end }}  
```

internal/ovncontroller/utils.go
```diff
@@ -18,12 +18,24 @@ import (
...
 	"golang.org/x/exp/maps"
 	corev1 "k8s.io/api/core/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+// ComputeSystemID derives a deterministic UUID5 system-id from a node name,
+// matching the convention used by OVN for chassis identification.
+func ComputeSystemID(nodeName string) string {
+	return uuid.NewSHA1(uuid.NameSpaceDNS, []byte(nodeName)).String()
+}
```
stuggi, line 32: is it required to use  UUID5 system-id ? wouldn't it be easier for debugging if the nodeNames are used/shown?

stuggi: the current edpm nodes ovn certs already have `CN=hostname`. if we'd use the same CA and nodename as CN it might make transition/migration easier?

```
        Subject: O=rootca-ovn, CN=compute-n95n1war-0
```

internal/ovncontroller/cert.go
```diff
@@ -0,0 +1,115 @@
...
+	"k8s.io/apimachinery/pkg/types"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+// EnsureRbacCert creates or updates a cert-manager Certificate CR for a node's
+// RBAC client certificate, and waits for the resulting Secret to become ready.
+func EnsureRbacCert(
```
stuggi: I think we could use EnsureCert from https://github.com/openstack-k8s-operators/lib-common/blob/main/modules/certmanager/certificate.go#L172 ?

using the cert-duration from the issuer annotation https://github.com/openstack-k8s-operators/lib-common/blob/main/modules/certmanager/issuer.go#L169:

```
apiVersion: cert-manager.io/v1
kind: Issuer
metadata:
  annotations:
    cert-duration: 43800h0m0s
```

internal/ovndbcluster/statefulset.go
```diff
@@ -133,6 +133,30 @@ func StatefulSet(
...
 		volumes = append(volumes, svc.CreateVolume(serviceName))
 		volumeMounts = append(volumeMounts, svc.CreateVolumeMounts(serviceName)...)
+
+		// For SB DB, mount the RBAC CA cert so we can build a combined CA bundle
+		if instance.Spec.DBType == ovnv1.SBDBType && instance.Spec.RbacCACertSecretName != "" {
+			volumes = append(volumes, corev1.Volume{
+				Name: "rbac-ca-cert",
+				VolumeSource: corev1.VolumeSource{
+					Secret: &corev1.SecretVolumeSource{
+						SecretName: instance.Spec.RbacCACertSecretName,
```
stuggi: iiuc, OVN RBAC checks CN vs system-id, not CA identity. Using the existing rootca-ovn to sign RBAC certs would eliminate: the  RbacCACertSecretName field, the RBAC CA volume mount on SB pods, and the combined bundle construction in setup.sh.

templates/ovndbcluster/bin/setup.sh
```diff
@@ -17,8 +17,15 @@ set -ex
...
 DB_PORT="{{ .DB_PORT }}"
+DB_PORT_FULL_ACCESS="{{ .DB_PORT_FULL_ACCESS }}"
 {{- if .TLS }}
 DB_SCHEME="pssl"
+{{- if index . "OVN_RBAC_CACERT_PATH" }}
+SSL_CA_CERT=/etc/ovn/combined-ca.pem
+cat {{.OVNDB_CACERT_PATH}} {{ index . "OVN_RBAC_CACERT_PATH" }} > ${SSL_CA_CERT}
+{{- else }}
+SSL_CA_CERT={{.OVNDB_CACERT_PATH}}
+{{- end }}
```
stuggi: not required if we'd use the dedicated already existing ovn CA

internal/ovncontroller/cert.go
```diff
@@ -0,0 +1,115 @@
...
+) (ctrl.Result, error) {
+	Log := log.FromContext(ctx).WithName("Controllers").WithName("OVNController")
+
+	cert := &certmgrv1.Certificate{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      certName,
+			Namespace: instance.Namespace,
+			Labels:    labels,
+		},
+		Spec: certmgrv1.CertificateSpec{
```
stuggi: we are not setting a duration for the cert. maybe it should be longer then the default, like we do with other service certs.

internal/ovncontroller/cert.go
```diff
@@ -0,0 +1,115 @@
...
+	if err != nil {
+		if !k8s_errors.IsNotFound(err) {
+			return ctrl.Result{}, err
+		}
+		err = controllerutil.SetControllerReference(instance, cert, scheme)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+		Log.Info("Creating RBAC certificate", "name", certName, "cn", commonName)
+		err = c.Create(ctx, cert)
```
stuggi: one thing which can be followed up is cleanup of RBAC certs for no longer existing nodes?

stuggi (commented):
I don't think a dedicated rootca-ovn-rbac CA is needed here. OVN RBAC only checks that the client cert CN matches the chassis system-id — it does not care which CA signed the cert. The existing rootca-ovn CA can sign the RBAC certs just fine, which would remove the need for the new CA, the combined CA bundle in setup.sh, and the RbacCACertSecretName field. Or am I missing a point?

This would also make the EDPM side much simpler: the existing per-node EDPM OVN cert is already signed by rootca-ovn and has CN=<hostname> with client auth usage, so it already works for RBAC as-is. The only thing needed on the EDPM side would be to set external-ids:system-id to the hostname so it matches the cert CN, instead of letting ovn-controller generate a random UUID. No new certs or CA changes needed for EDPM.

Also, I think using CN=nodeName instead of CN=UUID5(nodeName) would be simpler and easier to debug — ovn-sbctl list chassis would show actual node names. 

What you think?

[update]
initially I was wondering if we could just do the rbac cert creation in the openstack-op for the ovn-controllers and kust provide the list of cert secrets. but then realized that the ovn-controllers use daemonset (was thinking its a statefulset or so), which makes it harder to predict which nodes it gets scheduled to from there, especially if you use nodeselector, or topologyRef. therefor its probably ok to issue the rbac certs for the  ovn-controller k8s side in the operator.

internal/ovncontroller/daemonset.go
slawqo: My idea was to make it uncoditional and use it always when TLS is enabled in the environment. But we can add knob to enable/disable it if needed

internal/ovndbcluster/service.go
slawqo: same answer here - my idea is to make it always enabled when TLS is enabled

internal/ovncontroller/utils.go
slawqo: My initial idea was to use hostnames but the problem is that we would need to make changes in neutron for that as today it simply use what is in the system-id and use it as Neutron agent's ID. But more I think about it more I think that I should deeper check it in Neutron

templates/ovndbcluster/bin/setup.sh
slawqo: as I wrote above, my idea is to use RBAC always when TLS is enabled

internal/ovndbcluster/statefulset.go
slawqo: You are right, I am going to change it. Patches will be smaller then

templates/ovndbcluster/bin/setup.sh
slawqo: yes, will be changed

internal/ovncontroller/cert.go
slawqo: done

internal/ovncontroller/cert.go
slawqo: it is changed now

internal/ovncontroller/cert.go
slawqo: I will add new task for this in the epic and will work on it once the initial patches will be done. Is that ok for you?

slawqo: I've just created task for that https://redhat.atlassian.net/browse/OSPRH-29979

internal/controller/ovncontroller_controller.go
```diff
@@ -641,6 +648,26 @@ func (r *OVNControllerReconciler) reconcileNormal(ctx context.Context, instance
...
+	// permission tables are populated by ovn-northd, so ovn-controller must
+	// not start until northd is running; otherwise all SB operations are
+	// denied with "permission error".
+	if instance.Spec.TLS.Enabled() {
+		northd, err := ovnv1.GetOVNNorthd(ctx, helper, instance.Namespace)
+		if err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to look up OVNNorthd: %w", err)
+		}
+		if northd == nil || northd.Status.ReadyCount == 0 {
+			Log.Info("OVNNorthd is not ready yet, waiting before deploying OVNController...")
```
karelyatin, line 661: will check more tomorrow, but this got my attention and got curios on the impact on minor updates and adoption when tls is enabled?
like till now we allow updates to process in multiple maintenance windows. After this change will that stop work or /me missing something here? https://docs.redhat.com/en/documentation/red_hat_openstack_services_on_openshift/18.0/html/updating_your_environment_to_the_latest_maintenance_release/assembly_preparing-for-a-minor-update_update-rhoso#exit-the-update-process-if-an-issue-occurs_preparing-minor-update

Also this seems will impact even without minor update and just with operator updates. i.e with operator updates ovn operator reconcile RBAC changes will get applied on control plane side, and with that OVN controllers on dataplane will not be able to connect until minor update is done?

Also in case of OVN agent which also connects with NB DB, is there some issue on doing with NB? or that will be taken care separately?

PR #570, opened 2026-05-07
Add AGENTS.md

fmount:
This patch adds `AGENTS.md` to the repo to ensure any `AI agent` can follow the conventions defined for this operator. At the same time, it allows to not waste time navigating the repo to find initial information and save tokens.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,100 @@
+# AGENTS.md - ovn-operator
+
+## Project overview
+
+ovn-operator is a Kubernetes operator that manages
+[OVN](https://www.ovn.org/) (Open Virtual Network: distributed virtual
+networking for OpenStack including northd, OVN DB clusters, and OVN
+controllers) on OpenShift/Kubernetes. It is part of the
+[openstack-k8s-operators](https://github.com/openstack-k8s-operators) project.
+
+Key OVN domain concepts: **OVN Northd** (central daemon translating logical
+to physical flows), **OVN DB clusters** (northbound and southbound OVSDB
+clusters), **OVN Controller** (per-node agent managing local OVS flows),
+**OVS external IDs** (Open vSwitch configuration), **bonds** (NIC bonding).
+
+## Tech stack
+
+| Layer | Technology |
+|-------|------------|
+| Language | Go (modules, multi-module workspace via `go.work` - local only, not committed) |
+| Scaffolding | [Kubebuilder v4](https://book.kubebuilder.io/) + [Operator SDK](https://sdk.operatorframework.io/) |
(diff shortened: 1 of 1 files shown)


AGENTS.md
```diff
@@ -0,0 +1,94 @@
...
+- KUTTL integration tests live in `test/kuttl/`.
+- Run all functional tests: `make test`.
+- When adding a new field or feature, add corresponding test cases in
+  `test/functional/` and update fixture data accordingly.
+
+## Key dependencies
+
+- [lib-common](https://github.com/openstack-k8s-operators/lib-common): shared modules for conditions, endpoints, TLS, secrets, etc.
+- [infra-operator](https://github.com/openstack-k8s-operators/infra-operator): topology APIs.
+- [dev-docs/developer.md](https://github.com/openstack-k8s-operators/dev-docs/blob/main/developer.md): developer guide and coding conventions.
```
slawqo, line 100: should we also mention neutron-operator here?

fmount: yeah makes sense, let me update this and also mention pre-commit like we did in neutron-operator.

fmount: (Done)

karelyatin: But neutron-operator is dependent on the ovn-operator not the reverse so seems we don't need to add here, if we add then it will also require octavia-operator.
It is something like almost all operator depends on keystone-operator, and in keystone-operator all operators needs to be added as dependency which doesn't look so here checking keystone-operator similar PR.

AGENTS.md
```diff
@@ -0,0 +1,101 @@
...
+| Linting | golangci-lint (`.golangci.yaml`) |
+| CI | Zuul (`zuul.d/`), Prow (`.ci-operator.yaml`), GitHub Actions |
+
+## Custom Resources
+
+| Kind | Purpose |
+|------|---------|
+| `OVNNorthd` | Manages the OVN Northd deployment (logical-to-physical flow translation). |
+| `OVNDBCluster` | Manages OVN database clusters (northbound and southbound OVSDB). |
+| `OVNController` | Manages the OVN Controller daemonset (per-node OVS flow management). |
```
karelyatin: here we should extend ovs and metric daemonsets as well

AGENTS.md
fmount: ah ok, yes, if the dependency is the reverse we don't need to overcomplicate. for deeper inspection I assume the LLM can simply go through go.mod and build the dependency graph, but here is mostly to give a direction about the key dependency we want to establish for the project.
So yes, let's not have other dependencies here other than what we added at the beginning, and then we can improve in a follow up in case we missed something.

karelyatin (approved):
/lgtm

PR #573, opened 2026-05-20
Fix OVN encap IP detection

karelyatin:
With the current detection logic based on how
cluster is setup we can have multiple IPs detected while we expect single IP for this setting.
OVN itself support multiple IPs comma seperated and setup multiple tunnels for those and also require
to set default encap IP along.

But in our case we only require single IP. So to
fix this we now do as follows:-
- check for IPv4 address
- if IPv4 absent we fallback to IPv6
- if both are absent we exit
- If more than one IP exist for a type we choose first


karelyatin:
recheck mirror issue fixed

slawqo (approved):

PR #578, opened 2026-06-16
Move runtime config reconcilation post service setup

karelyatin:
We need to wait for ovn cluster to be ready and that relies on pods to be in ready state. And for multi replica ovn cluster the pods needs to communicate with each other using dns addresses and those are provided by the headless service.

So moving the runtime config task post setting up the services so cluster can be formed and pods move to the ready state. This issue was catched in the backup/restore exercise [2].

[1] https://github.com/openstack-k8s-operators/ovn-operator/pull/540
[2] https://github.com/openstack-k8s-operators/dev-docs/blob/main/backup-restore/user-guide.md

Related-Issue: #[OSPRH-27985](https://redhat.atlassian.net/browse/OSPRH-27985)

### DIFF
--- a/internal/controller/ovndbcluster_controller.go
+++ b/internal/controller/ovndbcluster_controller.go
@@ -659,24 +659,29 @@ func (r *OVNDBClusterReconciler) reconcileNormal(ctx context.Context, instance *
 		return ctrlResult, nil
 	}
 
+	// create Statefulset - end
+	// Handle service init
+	ctrlResult, err = r.reconcileServices(ctx, instance, helper, serviceLabels, serviceName)
+	if err != nil {
+		instance.Status.Conditions.Set(condition.FalseCondition(
+			condition.ExposeServiceReadyCondition,
+			condition.ErrorReason,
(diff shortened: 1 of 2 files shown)


internal/controller/ovndbcluster_controller.go
```diff
@@ -1005,7 +1013,7 @@ func (r *OVNDBClusterReconciler) reconcileServices(
 				return ctrl.Result{}, err
 			}
 
-			dnsIP, err := getPodIPInNetwork(ovnPod, instance.Namespace, instance.Spec.NetworkAttachment)
+			dnsIP, err := getPodIPInNetwork(ovnPod, instance)
 			dnsIPsList = append(dnsIPsList, dnsIP)
 			if err != nil {
```
stuggi, line 1017: should this error check happen before the above append?

karelyatin: yes that should be cleaner, updated.

stuggi (approved):
/lgtm

PR #584, opened 2026-06-26
Copy LSP.tag to LSP.tag_request if it is unset

otherwiseguy:
OVN 26.03.1 contains a fix that clears tags when the tag_request is cleared. For years, neutron has set the tag directly without setting a request. This if fixed in the neutron code, but on upgrade, it is possible that ports with tags and no tag_request will have the tags cleared before neutron could restart to fix them. Due to upgrade races between ovndbcluster and northd, this hacks the fix into northd startup.

Resolves: [OSPRH-31613](https://redhat.atlassian.net/browse/OSPRH-31613)

### DIFF
--- a/internal/ovnnorthd/statefulset.go
+++ b/internal/ovnnorthd/statefulset.go
@@ -31,7 +31,11 @@ import (
 
 const (
 	// ServiceCommand -
-	ServiceCommand = "/usr/bin/ovn-northd"
+	// TODO(otherwiseguy) Revert to /usr/bin/ovn-northd and remove setup.sh script
+	// once we are sure tag to tag_request migration no longer needed for a release
+	// i.e migration is already done in old running release and is the minimum required
+	// to move to the next release
+	ServiceCommand = "/usr/local/bin/container-scripts/setup.sh"
 )
 
 // StatefulSet func
--- a/templates/ovnnorthd/bin/setup.sh
+++ b/templates/ovnnorthd/bin/setup.sh
@@ -1,0 +1,42 @@
+#!/bin/bash
+set -e
+
+for arg in "$@"; do
+    case $arg in
+        --ovnnb-db=*) NB_DB="${arg#*=}" ;;
+        --certificate=*) CERT="${arg#*=}" ;;
+        --private-key=*) PKEY="${arg#*=}" ;;
+        --ca-cert=*) CACERT="${arg#*=}" ;;
+    esac
+done
+
+if [ -n "$NB_DB" ]; then
+    NBCTL="ovn-nbctl --no-leader-only"
+    if [ -n "$CERT" ]; then
+        NBCTL="$NBCTL --certificate=$CERT --private-key=$PKEY --ca-cert=$CACERT"
+    fi
+
+    # NOTE(twilson) This will wait for the ovsdb connection before detaching
+    export OVN_NB_DAEMON=$(${NBCTL} --pidfile --detach --db=$NB_DB)
+
+    # NOTE(twilson) ovn-nbctl will be connecting to the daemon successfully, so will wait forever
+    # until the daemon responds. So even if there is a disconnect after initially connecting it waits.
+    #
+    # Due to OVN commit `171ed24dd northd: Clear stale LSP tags on tag_request removal.` and neutron's
+    # historical practice of setting the tag directly instead of using tag_request, 26.03.1+ versions
+    # of northd will clear all tags set by neutron prior to 18.0 FR6. For upgrades to FR6 and beyond
+    # from an affected prior version (<= 18.0 FR6) this will fix the issue before northd starts.
+    #
+    # This is safe to remove when the minimum version from which upgrades are allowed is 18.0 FR7 or
+    # later. Note that this setup script itself only exists to run this fix.
+    for port in $(${NBCTL} --bare --columns _uuid find logical_switch_port tag!='[]' tag_request='[]'); do
+        tag=$(${NBCTL} lsp-get-tag $port)
+        echo "Fixing tag_request for $port tag=$tag"
+        ${NBCTL} set logical_switch_port $port tag_request=$tag
+    done
+
+    kill $(cat $OVN_RUNDIR/ovn-nbctl.pid)
+    unset OVN_NB_DAEMON
+fi
+
+exec /usr/bin/ovn-northd "$@"
--- a/test/kuttl/common/assert_sample_deployment.yaml
+++ b/test/kuttl/common/assert_sample_deployment.yaml
@@ -138,7 +138,7 @@ spec:
             weight: 100
       containers:
       - command:
-        - /usr/bin/ovn-northd
+        - /usr/local/bin/container-scripts/setup.sh
         imagePullPolicy: IfNotPresent
         livenessProbe:
           exec:


templates/ovnnorthd/bin/setup.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+set -e
```
karelyatin, line 2: ok this will exit on errors, but may be better if we also have some wait for ovndbs on this script startup?
considering the case where both ovndbs(single replica setup) and ovn-northd started or rolled out together.

karelyatin:
In northd logs https://gateway-cloud-softwarefactory.apps.ocp.cloud.ci.centos.org/logs//558/rdoproject.org/558bdff04a364d57a3cb4bbcbf95a0a9/controller/ci-framework-data/logs/openstack-must-gather/quay-io-openstack-k8s-operators-openstack-must-gather-sha256-60f18c13dd16095f37313b68f86965210c8ccdb0badbf530f299f3accfed406a/namespaces/openstack/pods/ovn-northd-0/logs/ovn-northd.log observed the below warning:-
2026-06-26T16:42:15Z|00002|ovn_dbctl|WARN|not using ovn-nbctl daemon because of db option

templates/ovnnorthd/bin/setup.sh
otherwiseguy: Sorry, I should have commented on that in the commit message. Using the nbctl daemon does cause it to wait on the connection.

otherwiseguy:
> In northd logs https://gateway-cloud-softwarefactory.apps.ocp.cloud.ci.centos.org/logs//558/rdoproject.org/558bdff04a364d57a3cb4bbcbf95a0a9/controller/ci-framework-data/logs/openstack-must-gather/quay-io-openstack-k8s-operators-openstack-must-gather-sha256-60f18c13dd16095f37313b68f86965210c8ccdb0badbf530f299f3accfed406a/namespaces/openstack/pods/ovn-northd-0/logs/ovn-northd.log observed the below warning:- 2026-06-26T16:42:15Z|00002|ovn_dbctl|WARN|not using ovn-nbctl daemon because of db option

@karelyatin Ah, good catch. I moved the --db option to only the nbctl daemon setup.

internal/ovnnorthd/statefulset.go
```diff
@@ -31,7 +31,7 @@ import (
 
 const (
 	// ServiceCommand -
-	ServiceCommand = "/usr/bin/ovn-northd"
+	ServiceCommand = "/usr/local/bin/container-scripts/setup.sh"
```
slawqo, line 38: IIUC the problem it is solving, this is needed only temporary and we should be fine with removing it in one of the next releases. We actually should need it only in 18, maybe even in one FR release and then it is not needed anymore, right? If that's correct understanding, maybe you could add some small TODO note to remove it in future, wdyt?

karelyatin: Actually we don't have control from which release to which release customer will update so atleast for 18.0 looks we need this for 18.0 Lifetime. i.e if any release before this fix is the source version then this will be required. For any future update(i.e once all the tag are switched to tag_request) it will not be needed

For 19 we likely would not need it depending on how the pre-requisite of rhoso 19 are layered out(like if pre-requisite is env should be on latest FR like FR7 then we should be fine to drop it), but if update is allowed from FR6 to RHOSO 19 then we can't get rid of it.
If we are sure of this we can push this patch only in 18.0-fr6 branch which will be the base of all FR and bug fix releases for 18.0
But yes good to add NOTE on the code why this all is needed with a TODO to remove once this is no longer needed

otherwiseguy:
I don't think the failure is related to this patch, Claude thinks it's a race condition but I have not verified:

```
● The PR is not causing this failure. Here's the analysis:

  What the PR changes (commit 607960c):
  1. internal/ovnnorthd/statefulset.go — changes the northd ServiceCommand from /usr/bin/ovn-northd to /usr/local/bin/container-scripts/setup.sh
  2. templates/ovnnorthd/bin/setup.sh — adds a new script that copies LSP.tag to LSP.tag_request before exec'ing northd

  What failed:
  All 10 kuttl tests failed because the operator controller manager service had no endpoints — the webhook service (movnnorthd-v1beta1.kb.io, movncontroller-v1beta1.kb.io, etc.) was unreachable. Two tests (ovn_config: 600s, ovn_readiness: 766s) hit timeouts; the remaining 8 failed instantly on webhook calls.

  Why the PR can't cause this:
  - The PR only changes what command the northd StatefulSet pods run — it doesn't touch the operator controller manager at all
  - The operator image built successfully (confirmed in build logs)
  - The ServiceCommand constant is used when the operator reconciles and creates the northd StatefulSet spec — it doesn't execute during operator startup
  - The setup.sh template gets loaded into a ConfigMap mounted into northd containers, not the operator container
  - No Go logic changes, no test references to the changed constant

  Root cause: This is a CI infrastructure flake — the operator controller manager pod either wasn't scheduled or crashed for reasons unrelated to this PR. The build log shows the operator deployment was initially "NotFound" (normal during OLM installation) and eventually reached REPLICAS=1, but something caused it to lose its endpoint later. This pattern (all kuttl tests failing on webhook endpoints) is a known CI transient failure mode.

✻ Cooked for 4m 23s

※ recap: You're working on the ovn-operator PR for OSPRH-31613 that adds a northd startup script to fix LSP tag_request on upgrade. The CI kuttl failure is an infrastructure flake unrelated to your change — safe to rerun. (disable recaps in /config)

❯ Can you tell what might need to happen to fix this flaky CI?                                                                                                                                                                                                                                                                                                                                                                                                                                

  Thought for 4s (ctrl+o to expand)

● Let me look at the kuttl test setup to understand how the operator deployment and webhook readiness are handled.

● Explore(Explore kuttl test infrastructure)
  ⎿  Done (62 tool uses · 106.6k tokens · 15m 46s)
  (ctrl+o to expand)

● Good analysis from the exploration. The core issue is a race condition: kuttl tests start applying CRs before the operator's webhook endpoints are ready to serve traffic.

  There are a few things that could fix this:

  1. Add a webhook readiness gate to the kuttl test setup — The most direct fix. Add an initial kuttl test step (or a script in the common deploy step) that waits for the webhook service to have endpoints before applying any CRs:
  oc wait --for=jsonpath='{.subsets[0].addresses}' \
    endpoints/ovn-operator-controller-manager-service \
    -n openstack-operators --timeout=120s

  2. Change failurePolicy from Fail to Ignore — This would let CR creation succeed even when webhooks are unavailable, but that's a bad idea for production since it bypasses validation.

  3. Add retry logic to the deploy step — The kuttl assert steps already retry, but the apply steps (which create the CRs and trigger webhooks) don't. Wrapping the CR creation in a retry loop in the deploy script would handle transient webhook unavailability.

  4. Fix the readiness probe to include webhook port — The operator's readiness probe checks port 8081 (/readyz), but the StartedChecker from mgr.GetWebhookServer() is registered on that same health endpoint. If the webhook server hasn't started yet, the readiness probe should fail — so the issue may be that the Kubernetes endpoints controller briefly adds the pod to the service before the readiness probe gates it out.

  The most practical fix is option 1 — it's non-invasive, doesn't change production behavior, and directly addresses the race. But since this CI infrastructure likely lives in the openstack-k8s-operators-ci shared repo (where the build/deploy/kuttl orchestration happens), the fix might need to go there rather than in ovn-operator itself. Want me to look at where exactly in the CI pipeline that wait should be inserted?

```

karelyatin:
> I don't think the failure is related to this patch, Claude thinks it's a race condition but I have not verified:
> 
> ```
> ● The PR is not causing this failure. Here's the analysis:
> 
>   What the PR changes (commit 607960c):
>   1. internal/ovnnorthd/statefulset.go — changes the northd ServiceCommand from /usr/bin/ovn-northd to /usr/local/bin/container-scripts/setup.sh
>   2. templates/ovnnorthd/bin/setup.sh — adds a new script that copies LSP.tag to LSP.tag_request before exec'ing northd
> 
>   What failed:
>   All 10 kuttl tests failed because the operator controller manager service had no endpoints — the webhook service (movnnorthd-v1beta1.kb.io, movncontroller-v1beta1.kb.io, etc.) was unreachable. Two tests (ovn_config: 600s, ovn_readiness: 766s) hit timeouts; the remaining 8 failed instantly on webhook calls.
> 
>   Why the PR can't cause this:
>   - The PR only changes what command the northd StatefulSet pods run — it doesn't touch the operator controller manager at all
>   - The operator image built successfully (confirmed in build logs)
>   - The ServiceCommand constant is used when the operator reconciles and creates the northd StatefulSet spec — it doesn't execute during operator startup
>   - The setup.sh template gets loaded into a ConfigMap mounted into northd containers, not the operator container
>   - No Go logic changes, no test references to the changed constant
> 
>   Root cause: This is a CI infrastructure flake — the operator controller manager pod either wasn't scheduled or crashed for reasons unrelated to this PR. The build log shows the operator deployment was initially "NotFound" (normal during OLM installation) and eventually reached REPLICAS=1, but something caused it to lose its endpoint later. This pattern (all kuttl tests failing on webhook endpoints) is a known CI transient failure mode.
> 
> ✻ Cooked for 4m 23s
> 
> ※ recap: You're working on the ovn-operator PR for [OSPRH-31613](https://redhat.atlassian.net/browse/OSPRH-31613) that adds a northd startup script to fix LSP tag_request on upgrade. The CI kuttl failure is an infrastructure flake unrelated to your change — safe to rerun. (disable recaps in /config)
> 
> ❯ Can you tell what might need to happen to fix this flaky CI?                                                                                                                                                                                                                                                                                                                                                                                                                                
> 
>   Thought for 4s (ctrl+o to expand)
> 
> ● Let me look at the kuttl test setup to understand how the operator deployment and webhook readiness are handled.
> 
> ● Explore(Explore kuttl test infrastructure)
>   ⎿  Done (62 tool uses · 106.6k tokens · 15m 46s)
>   (ctrl+o to expand)
> 
> ● Good analysis from the exploration. The core issue is a race condition: kuttl tests start applying CRs before the operator's webhook endpoints are ready to serve traffic.
> 
>   There are a few things that could fix this:
> 
>   1. Add a webhook readiness gate to the kuttl test setup — The most direct fix. Add an initial kuttl test step (or a script in the common deploy step) that waits for the webhook service to have endpoints before applying any CRs:
>   oc wait --for=jsonpath='{.subsets[0].addresses}' \
>     endpoints/ovn-operator-controller-manager-service \
>     -n openstack-operators --timeout=120s
> 
>   2. Change failurePolicy from Fail to Ignore — This would let CR creation succeed even when webhooks are unavailable, but that's a bad idea for production since it bypasses validation.
> 
>   3. Add retry logic to the deploy step — The kuttl assert steps already retry, but the apply steps (which create the CRs and trigger webhooks) don't. Wrapping the CR creation in a retry loop in the deploy script would handle transient webhook unavailability.
> 
>   4. Fix the readiness probe to include webhook port — The operator's readiness probe checks port 8081 (/readyz), but the StartedChecker from mgr.GetWebhookServer() is registered on that same health endpoint. If the webhook server hasn't started yet, the readiness probe should fail — so the issue may be that the Kubernetes endpoints controller briefly adds the pod to the service before the readiness probe gates it out.
> 
>   The most practical fix is option 1 — it's non-invasive, doesn't change production behavior, and directly addresses the race. But since this CI infrastructure likely lives in the openstack-k8s-operators-ci shared repo (where the build/deploy/kuttl orchestration happens), the fix might need to go there rather than in ovn-operator itself. Want me to look at where exactly in the CI pipeline that wait should be inserted?
> ```

Claude is wrong here :) test failures are related to this patch, https://github.com/openstack-k8s-operators/ovn-operator/blob/0d0d224f6f3390dffca4dc133923ae5a6e44d8cf/test/kuttl/common/assert_sample_deployment.yaml#L141 needs to be fixed

internal/ovnnorthd/statefulset.go
karelyatin: Added a TODO for future cleanup

karelyatin (approved):
/lgtm

PR #587, opened 2026-07-01
[ovncontroller] Fix ready conditions

karelyatin:
Currently we were relying only on daemonset
status DesiredNumberScheduled but we should
also consider daemonset running with updated
version i.e consider UpdatedNumberScheduled
as that is important to check during minor
updates.

Enhanced readyness condition so consider
all these factors.

Resolves: https://redhat.atlassian.net/browse/OSPRH-31318
Assisted-By: Claude

### DIFF
--- a/internal/controller/ovncontroller_controller.go
+++ b/internal/controller/ovncontroller_controller.go
@@ -714,22 +714,11 @@ func (r *OVNControllerReconciler) reconcileNormal(ctx context.Context, instance 
 	}
 	instance.Status.Conditions.MarkTrue(condition.NetworkAttachmentsReadyCondition, condition.NetworkAttachmentsReadyMessage)
 
-	// Check if all DaemonSets are ready
-	allReady := instance.Status.NumberReady == instance.Status.DesiredNumberScheduled &&
-		instance.Status.OVSNumberReady == instance.Status.DesiredNumberScheduled
-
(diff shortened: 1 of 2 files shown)


averdagu:
Code looks good to me, waiting for tests to give approval

karelyatin:
unrelated failure

internal/controller/ovncontroller_controller.go
```diff
@@ -1011,6 +1024,36 @@ func (r *OVNControllerReconciler) createMetricsHashOfInputHashes(
...
+// isDaemonSetRolloutComplete checks if a DaemonSet rollout is complete and all pods are ready
+func (r *OVNControllerReconciler) isDaemonSetRolloutComplete(ds appsv1.DaemonSet) bool {
+	// If nothing is desired to be scheduled, consider rollout complete
+	if ds.Status.DesiredNumberScheduled == 0 {
+		return true
+	}
+
+	// Check if generation is observed (ensures we're looking at current spec)
+	if ds.Status.ObservedGeneration < ds.Generation {
+		return false
```
slawqo, line 1031: so if the desired number scheduled is `0` we can assume that it Rollout is completed even if we are looking for old generation? Shouldn't we maybe change those 2 conditions?

karelyatin: Done

slawqo (approved):

PR #609, opened 2026-07-29
[ovn-controller] Avoid restarting ovs when TLS config changes

averdagu:
Currently, out of the two daemonsets (ovn-controller and ovn-controller-ovs) only the first one needs to know if TLS is enabled or not, ovn-controller-ovs is agnostic to the TLS setting.

This happens due to the fact that both daemonsets uses the same configMapVars to calculate the
hash, both including TLS options even though ovn-controller-ovs doesn't need it.
This commit adds a configMap for ovn-controller and another one for ovn-controller-ovs.

Resolves: [OSPRH-33030](https://redhat.atlassian.net/browse/OSPRH-33030)

### DIFF
--- a/api/v1beta1/ovncontroller_types.go
+++ b/api/v1beta1/ovncontroller_types.go
@@ -17,18 +17,24 @@ limitations under the License.
 package v1beta1
 
 import (
-	"k8s.io/apimachinery/pkg/util/validation/field"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/condition"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/tls"
+	"k8s.io/apimachinery/pkg/util/validation/field"
 
+	topologyv1 "github.com/openstack-k8s-operators/infra-operator/apis/topology/v1beta1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	topologyv1 "github.com/openstack-k8s-operators/infra-operator/apis/topology/v1beta1"
 )
 
 const (
 	// OVNConfigHash - OVNConfigHash key
 	OVNConfigHash = "OvnConfigHash"
+
+	// OVNInputHash
+	OVNInputHash = "OvnInputHash"
+
+	// OVSInputHash
+	OVSInputHash = "OvsInputHash"
 
 	// Container image fall-back defaults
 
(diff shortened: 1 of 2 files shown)


internal/controller/ovncontroller_controller.go
```diff
@@ -990,7 +1022,7 @@ func (r *OVNControllerReconciler) createHashOfInputHashes(
 	if err != nil {
 		return hash, changed, err
 	}
-	if hashMap, changed = util.SetHash(instance.Status.Hash, common.InputHashName, hash); changed {
+	if hashMap, changed = util.SetHash(instance.Status.Hash, hashkey, hash); changed {
 		instance.Status.Hash = hashMap
 		Log.Info(fmt.Sprintf("Input maps hash %s - %s", common.InputHashName, hash))
```
averdagu: I see I missed changing the InputHashName, will do that on next patchset

api/v1beta1/ovncontroller_types.go
```diff
@@ -17,19 +17,25 @@ limitations under the License.
...
 import (
-	"k8s.io/apimachinery/pkg/util/validation/field"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/condition"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/tls"
+	"k8s.io/apimachinery/pkg/util/validation/field"
 
+	topologyv1 "github.com/openstack-k8s-operators/infra-operator/apis/topology/v1beta1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	topologyv1 "github.com/openstack-k8s-operators/infra-operator/apis/topology/v1beta1"
```
slawqo, line 26: why those 2 changes?

internal/controller/ovncontroller_controller.go
```diff
@@ -972,6 +985,24 @@ func (r *OVNControllerReconciler) generateServiceConfigMaps(
...
+
+// createHashOfInputsHashesOVS - calls create Hash of Hashes to modify the OVS input hash
+func (r *OVNControllerReconciler) createHashOfInputHashesOVS(
+	ctx context.Context,
+	instance *ovnv1.OVNController,
+	envVars map[string]env.Setter,
+) (string, bool, error) {
+	return r.createHashOfInputHashes(ctx, instance, envVars, ovnv1.OVSInputHash)
+}
+
```
slawqo: do we need those wrapper functions which are doing basically nothing new?

slawqo:
btw. can we maybe test this new behaviour in e.g. functional tests to make sure that ovn-controller-ovs POD is not touched when TLS config will change?

api/v1beta1/ovncontroller_types.go
averdagu: I think this is something that my IDE did automatically, I can revert it

internal/controller/ovncontroller_controller.go
averdagu: True, removing this two functions

averdagu:
> btw. can we maybe test this new behaviour in e.g. functional tests to make sure that ovn-controller-ovs POD is not touched when TLS config will change?

Yes, I think it would be valuable to modify the kuttl test, will do that :)

